PR #2, opened 2016-04-04
Back to good ol' Sonatype

cb372:
The world is not quite ready for Bintray. Let's go back to the tried and tested Maven Central via Sonatype method for publishing.

See https://github.com/sbt/sbt/issues/2464 for more details. Basically JCenter had some reliability problems, so it's disabled by default in sbt for now. Telling our users to enable it is quite a big ask.


LATaylor-guardian:
:crying_cat_face:

PR #3, opened 2016-04-05
Add VideoStatsResponse to CAPI thrift model

tomrf1:
Initially this is to support most-viewed videos data. In future we may add more info to this response.

### DIFF
--- a/src/main/thrift/content/v1.thrift
+++ b/src/main/thrift/content/v1.thrift
@@ -1148,6 +1148,15 @@ struct Package {
     3: required string packageName
 }
 
+struct MostViewedVideo {
+
+    /* The video ID */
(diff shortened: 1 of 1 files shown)


src/main/thrift/content/v1.thrift
```diff
@@ -1148,6 +1148,15 @@ struct Package {
     3: required string packageName
 }
 
+struct MostViewedVideo {
+
+    /* The video ID */
+    1: required string id
+
+    /* The view count */
+    2: required i32 video
```
LATaylor-guardian: Is `video` the best name for the view count for this video?

LATaylor-guardian:
:+1:

PR #5, opened 2016-04-13
Add AtomUsageResponse struct

cb372:
This is the response that shows where a given atom is being used.


LATaylor-guardian:
Good spot. 👍

PR #6, opened 2016-04-14
Add JSON parser as a separate project

tomrf1:
The JSON parsing code in the content-api-scala-client was recently duplicated in concierge.
It's better to have it as a separate project in content-api-models. So we now publish 2 projects:

- content-api-models
- content-api-models-json

A nice consequence of this is that any thrift file changes will now actually be validated during the build.


build.sbt
```diff
@@ -1,49 +1,109 @@
...
+    libraryDependencies ++= Seq(
+      "org.json4s" %% "json4s-native" % "3.3.0",
+      "org.json4s" %% "json4s-ext" % "3.3.0",
+      "joda-time" % "joda-time" % "2.3",
+      "net.databinder.dispatch" %% "dispatch-core" % "0.11.3",
+      "org.scalatest" %% "scalatest" % "2.2.1" % "test",
+      "com.google.guava" % "guava" % "19.0" % "test"
+    ),
+
+    mappings in (Compile, packageBin) ~= { _.filter( f => f._1.getAbsolutePath.contains("com/gu/contentapi/json"))}
```
cb372: This should be done for `packageSrc` as well, so we don't include the unnecessary files in the sources jar.

```
val jsonPackageFilter = (fp: (java.io.File, String)) => fp._1.getAbsolutePath.contains("com/gu/contentapi/json")

...

mappings in (Compile, packageBin) ~= { _.filter(jsonPackageFilter) },
mappings in (Compile, packageSrc) ~= { _.filter(jsonPackageFilter) }
```

For the javadoc jar, it's more complicated, as we can't just exclude files. The HTML in the javadoc will refer to files that we have excluded. 

I think we should just not publish a javadoc jar. It's not mandatory for Maven Central anyway.

```
publishArtifact in (Compile, packageDoc) := false
```

cb372:
👍 

Now we have tests, we should add a Travis config. I'll do that after this is merged.

PR #9, opened 2016-04-20
Use jackson instead of native json4s

tomrf1:
Hopefully the cause of the higher cpu usage on Concierge


cb372:
👍  once the build passes

PR #13, opened 2016-04-26
Implement Scrooge struct -> JSON serialization

cb372:
The json4s serialization code was originally written only for JSON parsing, i.e. Json -> Scrooge struct deserialization, and until now it has only been used this way. This PR makes a start on support for going the other way.

* Split the serialization codecs out of JsonParser into a new file
* Add tests for JSON serialization of Scrooge classes
* Add a field serializer to prevent the `_passthroughFields` field from showing up in the JSON
* Convert enum value names to hyphenated version when serializing to JSON
* Also clean up the JSON -> Scrooge deserialization code slightly
    * Use the "unknown" value when enum deserialization fails, instead of choosing an arbitrary value
    * Get rid of the JNull case, which was unnecessary. It was blindly copied from some StackOverflow post a long time ago.

Still to do in a separate PR: 
* Implement serialization of atoms
* Write tests for the remaining response types


json/src/main/scala/com/gu/contentapi/json/Serialization.scala
```diff
@@ -0,0 +1,233 @@
...
+    case JField("displayCredit", JString(s)) => JField("displayCredit", JBool(s.toBoolean))
+    case JField("legallySensitive", JString(s)) => JField("legallySensitive", JBool(s.toBoolean))
+    case JField("sensitive", JString(s)) => JField("sensitive", JBool(s.toBoolean))
+    case JField("explicit", JString(s)) => JField("explicit", JBool(s.toBoolean))
+    case JField("clean", JString(s)) => JField("clean", JBool(s.toBoolean))
+    case JField("safeEmbedCode", JString(s)) => JField("safeEmbedCode", JBool(s.toBoolean))
+  }
+
+  def thriftEnum2JString[A <: ThriftEnum]: PartialFunction[Any, JString] = {
+    case a: ThriftEnum => JString(pascalCaseToHyphenated(a.name))
```
tomrf1: what's the reason for this change in behaviour?

cb372: The reason is that the old behaviour was wrong :)

When CAPI renders an enum value in JSON, it writes it in lowercase hyphenated form (e.g. "newspaper-book-section"). And we need to maintain that behaviour so we don't break any clients who use JSON. But the equivalent Thrift enum name is e.g. `NewspaperBookSection`.

tomrf1:
👍

PR #15, opened 2016-05-04
Big step towards working JSON serialization of Content struct

cb372:
* Add helper methods to stringify parts of the JSON
* Customise the String repr of `Office` enum: unlike other enums, this one is uppercase
* Fix the `thriftEnum2JString` method
* Add a test for round-tripping an ItemResponse containing a piece of content
* Extend the `CapiDateTime` struct to preserve timezone information
* Update/fix some test data

Still to do in later PRs:

* Crosswords (nearly works, just need to tweak the date formats)
* Atoms (haven't looked at it yet, but it will probably be a nightmare)
* Tests for other kinds of ItemResponse (section, tag) and more complex variations of Content

### DIFF
--- a/json/src/main/scala/com/gu/contentapi/json/Serialization.scala
+++ b/json/src/main/scala/com/gu/contentapi/json/Serialization.scala
@@ -240,10 +240,16 @@ object Serialization {
 
   object DateTimeSerializer extends CustomSerializer[CapiDateTime](format => (
     {
-      case JString(s) => CapiDateTime.apply(ISODateTimeFormat.dateOptionalTimeParser().parseDateTime(s).getMillis)
+      case JString(s) =>
+        val dateTime = ISODateTimeFormat.dateOptionalTimeParser().withOffsetParsed().parseDateTime(s)
+        CapiDateTime.apply(dateTime.getMillis, dateTime.toString(ISODateTimeFormat.dateTime()))
     },
     {
-      case d: CapiDateTime => JString(new DateTime(d.dateTime).toString)
+      case d: CapiDateTime =>
+        val dateTime = ISODateTimeFormat.dateTime().withOffsetParsed().parseDateTime(d.iso8601)
+
+        // We don't include millis in JSON, because nearest-second accuracy is enough for anyone!
+        JString(dateTime.toString(ISODateTimeFormat.dateTimeNoMillis()))
     }
     ))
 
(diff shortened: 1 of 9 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -11,6 +11,14 @@ struct CapiDateTime {
      * Date times are represented as i64 - epoch millis
      */
     1: required i64 dateTime
+
+    /*
+     * Also represent as a string (yyyy-MM-dd`T`HH:mm:ss.SSSZZ) in order to preserve timezone info.
+     *
+     * Note: this field makes the i64 representation redundant.
+     */
+    2: required string iso8601
```
cb372: Note that I'm adding a new required field here. We'll need to upgrade `content-api-models` in CAPI and start sending the field in responses before we upgrade the models in the Scala client.

json/src/main/scala/com/gu/contentapi/json/Serialization.scala
```diff
@@ -69,8 +71,25 @@ object Serialization {
     case JField("safeEmbedCode", JString(s)) => JField("safeEmbedCode", JBool(s.toBoolean))
   }
 
-  def thriftEnum2JString[A <: ThriftEnum]: PartialFunction[Any, JString] = {
-    case a: ThriftEnum => JString(pascalCaseToHyphenated(a.name))
+  def thriftEnum2JString[A <: ThriftEnum](implicit classTag: ClassTag[A]): PartialFunction[Any, JString] = {
+    case a if classTag.runtimeClass.isInstance(a) => JString(pascalCaseToHyphenated(a.asInstanceOf[A].name))
+  }
```
tomrf1: Do you mind explaining what's going on here please? What was happening before? Where is the implicit parameter coming from?

cb372: Recap of what I explained offline:

* `case a: ThriftEnum` means the serializer will be used for all Thrift enums, not just those of type `A`. This is quite surprising behaviour of json4s, and it means that we can't customise the serialization of particular Thrift enums.
* We really want to write `case a: A`, but that doesn't work because the type `A` is erased at runtime.
* We work around it using `ClassTag`, which preserves compile-time class information to make it available at runtime.
* `ClassTag` is always available in implicit scope. The compiler supplies it for us.

tomrf1:
👍

PR #16, opened 2016-05-10
Adds paidContentType and aboutLink commercial fields

steppenwells:
Adds the new fields introduced in https://github.com/guardian/content-api/pull/1543 into the thrift model.


cb372:
👍

PR #17, opened 2016-05-13
Add lastModified field to Package model

tomrf1:


models/src/main/thrift/content/v1.thrift
```diff
@@ -1206,6 +1206,8 @@ struct Package {
 
     /* The package name */
     3: required string packageName
+
+    4: required string lastModified
```
cb372: Why not make it a `CapiDateTime`?

LATaylor-guardian: Is this okay to be required?

models/src/main/thrift/content/v1.thrift
```diff
@@ -1206,6 +1206,8 @@ struct Package {
 
     /* The package name */
     3: required string packageName
+
+    4: optional CapiDateTime lastModified
```
mchv: @tomrf1 I think this field is required, and you put it optional for compatibility?
In that case, @cb372 told me this is unnecessary.

tomrf1: So what should it be?
Elsewhere it is optional

LATaylor-guardian: @tomrf1 I was under the same impression as Mariot. So if @cb372 thinks it can be marked as required without causing compatibility issues, then it should be `required`.

cb372:
👍

PR #18, opened 2016-05-13
Add a new membership tiers for premium content

mchv:
The new membership tiers is to take in account the existing premium mode which enable paid members on mobile apps top get access to premium content.


mchv:
@clloyd we agreed there is a better way to do this

PR #20, opened 2016-05-13
Move events.thrift from Crier into content-api-models.

LATaylor-guardian:
Most clients have to rely on this project anyway when using Crier so it makes sense for the events.thrift file to live in here also. This also prevents it from being copied and posted everywhere.

### DIFF
--- a/models/src/main/thrift/events/crier/event.thrift
+++ b/models/src/main/thrift/events/crier/event.thrift
@@ -1,0 +1,34 @@
+namespace scala com.gu.crier.model.event.v1
+
+include "content/v1.thrift"
+
+enum EventType {
+    Update = 1,
+    Delete = 2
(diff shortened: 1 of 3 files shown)


models/src/main/thrift/events/crier/event.thrift
```diff
@@ -0,0 +1,36 @@
+namespace scala com.gu.crier.model.event.v1
+
+include "content/v1.thrift"
+
+typedef string RemovedContent
```
mchv: I don't remember why this is not linked to any event

LATaylor-guardian: It's just a layover and needs to be cleaned up. Will remove.

cb372:
👍

PR #21, opened 2016-05-16
Add a RetrievableContent payload type.

oilnam:
This is used to handle large update events in Crier

### DIFF
--- a/models/src/main/thrift/events/crier/event.thrift
+++ b/models/src/main/thrift/events/crier/event.thrift
@@ -2,22 +2,47 @@ namespace scala com.gu.crier.model.event.v1
 
 include "content/v1.thrift"
 
+/*
+ * For large update events the EventType is set to RetrievableUpdate
+ * and the EventPayload.Content is replaced by an EventPayload.RetrievableContent
+ */
 enum EventType {
+
     Update = 1,
-    Delete = 2
+    Delete = 2,
+    RetrievableUpdate = 3
 }
 
 enum ItemType {
+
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/events/crier/event.thrift
```diff
@@ -2,22 +2,47 @@ namespace scala com.gu.crier.model.event.v1
 
 include "content/v1.thrift"
 
+/*
+ * For large update events the EventType is set to RetrievableUpdate
+ * and the EventPayload.Content is replaced by an EventPayload.RetrievableContent
```
LATaylor-guardian: Not sure we need this line of the comment.

models/src/main/thrift/events/crier/event.thrift
```diff
@@ -2,22 +2,47 @@ namespace scala com.gu.crier.model.event.v1
 
 include "content/v1.thrift"
 
+/*
+ * For large update events the EventType is set to RetrievableUpdate
```
LATaylor-guardian: This can be inlined alongside `RetrievableUpdate = 3` below.

LATaylor-guardian:
👍

models/src/main/thrift/events/crier/event.thrift
oilnam: it's just a bit too long to be inlined, and I think we agreed of leaving a comment because it's not entirely obvious just by looking at the model why we're doing this (think people outside capi?)

PR #23, opened 2016-05-25
Add internal revision field to Crier's model

oilnam:
I should've merged this before releasing the new model, oops. Sorry.

### DIFF
--- a/models/src/main/thrift/events/crier/event.thrift
+++ b/models/src/main/thrift/events/crier/event.thrift
@@ -36,6 +36,11 @@ struct RetrievableContent {
      * The timestamp for when that specific payload was last modified
      */
     3: optional i64 lastModifiedDate
+
+    /*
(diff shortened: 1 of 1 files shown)


cb372:
Why not add the `internalRevision` field to the `RetrievableContent` struct instead? It only makes sense for content, not other types we might add in the future.

LATaylor-guardian:
Yes, It will need to be available in the `Content ` and `RetrievableContent`struct. In theory, nothing else should need to be added to the `Event` struct. It's purposely kept quite generic.

oilnam:
@LATaylor-guardian I see your point. I added it to `RetrievableContent` only, since we can already get it from inside `Content` (I added it in a previous PR).

cb372:
👍

LATaylor-guardian:
👍

PR #25, opened 2016-05-26
Add the Section field to Content

oilnam:
I had to move `struct Section` and `struct Edition` further up before `struct Content` because I'm now referencing `Section` from `Content`

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -971,6 +971,72 @@ struct Tag {
     19: optional string paidContentType
 }
 
+struct Edition {
+
+    /*
+     * he path of the edition, e.g. 'au/business'
+     */
+    1: required string id
+
+    /*
+     * Short description of the edition
+     */
+    2: required string webTitle
+
+    /*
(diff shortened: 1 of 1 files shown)


cb372:
👍

models/src/main/thrift/content/v1.thrift
```diff
@@ -1095,72 +1161,12 @@ struct Content {
...
-     * For sections, this allows access to the editorsPicks for the section,
-     * mostRead content in the section,
-     * and automatically shows the most recent content for the section.
-     */
-    4: required string apiUrl
-
-    /*
-     * List of available editions for this section
+     * The section associated with this content.
+     * Only returned if you specify showSection(true) on the request
```
LATaylor-guardian: This part of the comment is not needed. The Thrift definition is agnostic of how we choose for our parameters to behave in Concierge.

oilnam: I copied it from the comment we have on the Tag struct. I already merged unfortunately, but maybe we can get rid of both comments in the future

PR #27, opened 2016-06-14
Remove viewpoints atom type

tomrf1:
I'll put this in the next major release, along with the thriftification and media atom changes


models/src/main/thrift/content/v1.thrift
```diff
@@ -1119,9 +1119,7 @@ struct Section {
 struct Atoms {
     1: optional list<contentatom.Atom> quizzes
 
-    2: optional list<contentatom.Atom> viewpoints
-
-    3: optional list<contentatom.Atom> media
+    2: optional list<contentatom.Atom> media
```
cb372: Let's keep `media` as field number 3 and leave the viewpoints line commented out as a reminder that we should never reuse field number 2. Reusing field numbers is a big no-no in Thrift.

models/src/main/thrift/content/v1.thrift
```diff
@@ -1401,9 +1399,7 @@ struct ItemResponse {
     /* New story packages */
     20: optional list<Package> packages
 
-    21: optional list<contentatom.Atom> viewpoints
-
-    22: optional contentatom.Atom media
+    21: optional contentatom.Atom media
```
cb372: Same as above

cb372:
👍  (although I would have left a comment in the Thrift file documenting what the missing field was)

tomrf1:
We decided in the end to leave the viewpoints fields in the thrift model

PR #28, opened 2016-06-14
Fix the build

tomrf1:
This broke the concierge build, because content-api-models-json picked up the model class files


cb372:
👍

PR #32, opened 2016-06-27
Make Rights fields required, default to false

tomrf1:
This was a breaking change introduced in preparation for thriftification, but we've decided it's not worth the pain


LATaylor-guardian:
👍 Once build passes.

PR #33, opened 2016-06-30
Circe support

tomrf1:
In preparation for thriftification, we are migrating from json4s to circe for json processing. This is partly because circe is faster, but also because it is more actively maintained.

1. Added new deserializers for the various types, using circe instead of json4s
2. New macros lib, for the circe deserialization logic
3. Added diff tests and benchmarking for circe vs json4s

We've left the json4s code in this library for now, but it can be removed in future.


build.sbt
```diff
@@ -66,6 +66,18 @@ lazy val root = Project(id = "root", base = file("."))
     )
   )
 
+lazy val macros = Project(id = "content-api-models-macros", base = file("macros"))
+  .settings(commonSettings)
+  .settings(
+    description := "Macros",
+    libraryDependencies ++= Seq(
+      "io.circe" %% "circe-core" % "0.5.0-M2",
```
cb372: Maybe worth extracting the Circe version into a `val` so we can keep it consistent between the projects

json/src/main/scala/com/gu/contentapi/json/CirceSerialization.scala
```diff
@@ -0,0 +1,177 @@
...
+        quizzes <- getAtoms(c, AtomType.Quiz)
+        media <- getAtoms(c, AtomType.Media)
+      } yield {
+        Atoms(quizzes = quizzes, media = media)
+      }
+    }
+
+    private def getAtom(c: HCursor, atomData: AtomData): Decoder.Result[Atom] = {
+      for {
+        id <- c.downField("id").as[String]
```
cb372: `c.downField("foo").as[Bar]` can be written slightly more simply as `c.get[Bar]("foo")`

cb372:
👍  apart from a couple of minor pieces of cleanup.

I can explain the macros offline to anyone who's interested.

PR #38, opened 2016-07-04
Produce jar file of generated classes from models.

LATaylor-guardian:


build.sbt
```diff
@@ -95,6 +95,33 @@ lazy val models = Project(id = "content-api-models", base = file("models"))
...
+    // See: https://github.com/twitter/scrooge/issues/199
+    scroogeThriftSources in Compile ++= {
+      (scroogeUnpackDeps in Compile).value.flatMap { dir => (dir ** "*.thrift").get }
+    },
+    libraryDependencies ++= Seq(
+      "org.apache.thrift" % "libthrift" % "0.9.1",
+      "com.twitter" %% "scrooge-core" % "4.5.0"
+    )
+  )
+
```
cb372: Could the `json` project now depend on the `classes` project? That would mean we could get rid of the copy-pasted Scrooge stuff from the `json` project and avoid generating the same classes twice.

We would want to mark it as a `provided` dependency. I think you can write that as `.dependsOn(classes % "provided")`

Also don't forget to add `classes` to the `aggregates()` of the root project.

mchv:
May we improve the readme to describe the different jars we produce and for each explain when people should have a dependency on?
`classes` suffix is a bit weird as a name as a jar is supposed to contain `classes`, but I will not care too much if we have documentation explaining what each jar contain.

LATaylor-guardian:
@mchv Happy to change to a more appropriate name if you have any ideas?

README.md
```diff
@@ -10,3 +10,15 @@ $ sbt release
 
 will publish to [Maven Central](http://search.maven.org/) via Sonatype. You will need Sonatype credentials and a PGP key. It can take up to 2hrs to show up in search.
 
+## Information about built jars
+
+The content-api-models project builds the following jar files: 
+
+* content-api-models-scala - Scrooge generated class files based on the Thrift definitions of the content api models found in the `content-api-models` dependency. 
+
+* content-api-models-json - Json parsers and deserializers. Used internall by the content api to convert from Elasticsearch returned json to the Scrooge-generated Thrift classes. As a client you should never need to depend on this.
```
cb372: It's also used by the Scala client, so it's not only internal to Concierge. In other words, clients might depend on it, but probably only transitively.

cb372:
👍  but let's do a `publishLocal` tomorrow and quickly check that all the jars contain the things we expect.

mchv:
@LATaylor-guardian waiting for you to merge and then let's publish a `9.9.1` including #41 changes

LATaylor-guardian:
@mchv Will do so this morning.

LATaylor-guardian:
done a `jar -tf` on each of the jars after `publishLocal` and their contents look sensible to me. Merging.

PR #40, opened 2016-07-05
adds vine to thrift

tsop14:
- Adds vine to thrift. Definition in porter [here](https://github.com/guardian/content-api/blob/master/porter/src/main/thrift/flexible/content.thrift#L847).

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -626,7 +626,7 @@ struct VineElementFields {
 
     1: required string originalUrl
 
-    2: required string title;
+    2: required string title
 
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -569,49 +569,75 @@ struct EmbedElementFields {
...
 
-    10: optional i32 commentId; 
+    10: optional i32 commentId
 }
 
+struct VineElementFields {
+
+    1: required string originalUrl
+
+    2: required string title;
```
cb372: semicolon

cb372:
👍

PR #42, opened 2016-07-06
Remove mappings for json project.

LATaylor-guardian:
The `scala` project is now a provided dependency meaning we no longer need this code.


cb372:
👍

PR #43, opened 2016-07-06
Improve the Thrift struct Decoder macro to only use shapeless Lazy where necessary

cb372:
This is because use of Lazy incurs a runtime cost. We should only use it when we need it to hack around implicit resolution problems in the compiler.

Before:

```
[info] Benchmark                                        Mode  Cnt        Score       Error  Units
[info] JsonDecodeBenchmark.circe                        avgt   40   104592.251 ±  4303.060  ns/op
[info] JsonDecodeBenchmark.circeDeserializeEachContent  avgt   40  1751011.694 ± 93215.817  ns/op
```

After:

```
[info] Benchmark                                        Mode  Cnt        Score       Error  Units
[info] JsonDecodeBenchmark.circe                        avgt   40    95609.173 ±  1740.968  ns/op
[info] JsonDecodeBenchmark.circeDeserializeEachContent  avgt   40  1344281.162 ± 47547.969  ns/op
```

Also

* Add a benchmark for a very large, heavily nested response
* Tweak dependencies

### DIFF
--- a/macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
+++ b/macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
@@ -4,6 +4,7 @@ import scala.language.experimental.macros
 import scala.reflect.macros.blackbox
 import com.twitter.scrooge.{ThriftEnum, ThriftStruct}
 import io.circe.Decoder
+import shapeless.Lazy
 
 /**
   * Macros for Circe deserialization of various Scrooge-generated classes.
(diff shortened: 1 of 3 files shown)


cb372:
Ran the Concierge tests against this version and everything still seems to work.

macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
```diff
@@ -109,7 +127,7 @@ object CirceScroogeMacros {
     _root_.io.circe.Decoder.instance((cursor: _root_.io.circe.HCursor) => {
       cursor.focus.asString match {
         case _root_.scala.Some(value) =>
-          val withoutHyphens = org.apache.commons.lang3.StringUtils.remove(value, '-')
+          val withoutHyphens = _root_.org.apache.commons.lang3.StringUtils.remove(value, '-')
```
tomrf1: what happens without this?

mchv:
👏 for the results!
💯 for  the investigation work to improve the code

PR #45, opened 2016-07-14
Add short id as an internal field

mchv:


LATaylor-guardian:
I thought we decided on not adding this?

mchv:
@LATaylor-guardian Just discovered another tool [that do the extraction](https://github.com/guardian/facia-scala-client/pull/169) so we have now:
- Discussion updater lambda
- Discussion powwow
- Guardian frontend
- Facia client

So it seems to me that providing it as internal may be a good compromise

LATaylor-guardian:
Discussed with @mchv 👍

models/src/main/thrift/content/v1.thrift
```diff
@@ -984,6 +984,11 @@ struct ContentFields {
     40: optional i32 internalContentCode
 
     41: optional bool isLive
+    
+    /*
+     * A short id used as a key for discussion and save for later feature
+     */
+    42: optional string internalShortId
```
gidsg: Does this include the proceeding slash? as Discussion keys are like `/p/12345`

johnduffell:
🎉

models/src/main/thrift/content/v1.thrift
mchv: @gidsg yep this is the plan

PR #49, opened 2016-07-25
Circe macro: discard history when decoding

cb372:
This is a partial workaround for travisbrown/circe#329. Instead of calling `.get` directly on the `HCursor`, we perform the down-fielding on its underlying `Cursor`. This means that history is not preserved, resulting in fewer object allocations and less GC pressure.

The only downside to this is that we can't include the cursor history in error messages, but we don't really care about that.


LATaylor-guardian:
👍

PR #55, opened 2016-08-17
Add missing rcsId and contributorId to Tag model.

LATaylor-guardian:


cb372:
👍

Let's have filtering in concierge to only expose these 2 fields on the internal tier.

LATaylor-guardian:
> Let's have filtering in concierge to only expose these 2 fields on the internal tier.

@cb372 Why do we need to provide new behaviour? These fields existed before and are not new fields as such?

cb372:
Just because they are internal fields, only needed by RCS these days, and we don't want anyone else to start depending on them.

LATaylor-guardian:
> Just because they are internal fields, only needed by RCS these days, and we don't want anyone else to start depending on them.

Okay, will add.

PR #60, opened 2016-09-12
Thrift to circe encoders

tomrf1:
### Background
We've been migrating our json handling from json4s to circe.
We already have decoders, which turn circe json objects into circe. Concierge is using these.
This change adds encoders for going the other way.

### Why do we need encoders?
1. For building json responses in concierge, which some clients still need. At the moment, if a client requests format=json then concierge has to use json4s to do the conversion. We'd like to get rid of json4s from concierge as well.
2. It's nice for testing.
3. It would be nice to make the thrift encoder/decoder macros open source. There's a bit more to be done first.

### Notes
1. We were previously storing CapiDateTime with milliseconds, but dropping the milliseconds during deserialization. Now the encoder returns the datetime with milliseconds - this makes the code and the tests simpler.
2. I've replaced SerializationSpec.scala with CirceRoundTripSpec.scala. As well as switching from json4s to circe, there's a slightly philosophical shift because the test files no longer contain the json that concierge produces (e.g. no bools as strings). The tests in this library validate the circe encoders/decoders, and concierge should (and does) test any crazy stringifying/destringifying that it does before or after building the thrift.
3. I've removed all references to serialization/deserialization (which is what we called it with json4s) because I always get the terminology mixed up now that we have encoder/decoder as well.

The associated concierge change is on its way...


json/src/main/scala/com/gu/contentapi/json/CirceEncoders.scala
```diff
@@ -0,0 +1,113 @@
...
+  implicit val videoStatsResponseEncoder = Encoder[VideoStatsResponse]
+  implicit val atomsUsageResponseEncoder = Encoder[AtomUsageResponse]
+  implicit val removedContentResponseEncoder = Encoder[RemovedContentResponse]
+
+  def genDateTimeEncoder: Encoder[CapiDateTime] = Encoder.instance[CapiDateTime] { capiDateTime =>
+    Json.fromString(capiDateTime.iso8601)
+  }
+
+  /**
+    * TODO - I *will* write a pair of macros for encoding/decoding thrift union types, then delete all this stuff.
```
philwills: You might want to talk to @emma-p and I if you do try and do this as we recently did https://github.com/guardian/marley/pull/3/files which was a bit hairy in places.

tomrf1: @philwills good to know, thanks!

benchmarks/src/main/scala/benchmark/JsonDecodeBenchmark.scala
```diff
@@ -22,77 +23,48 @@ case class SearchResp(
...
-    (jvalue, circeJson)
+    circeJson
   }
 
-  val (contentJvalue: JValue, contentJson: Json) = loadJson("content-with-blocks.json")
-  val (searchResponseJvalue, searchResponseJson) = loadJson("search-response.json")
-  val (massiveContentsListJvalue, massiveContentsListJson) = loadJson("massive-contents-list.json")
+  val contentJson = loadJson("content-with-blocks.json")
+  val massiveContentsListJson = loadJson("massive-contents-list.json")
+  val contentOpt = contentJson.as[Content].leftMap(e => throw e).toOption
```
cb372: As this is just a benchmark, I would do a `.get` here. Seems weird to leave it as an option, and you don't want the time taken to extract the value to interfere with your benchmarking.

json/src/main/scala/com/gu/contentapi/json/CirceDecoders.scala
```diff
@@ -10,38 +10,8 @@ import com.gu.contentatom.thrift.atom.cta.CTAAtom
...
-      case JLong(value) => Json.fromLong(value)
-      case JDouble(value) => Json.fromDoubleOrNull(value)
-      case JDecimal(value) => Json.fromBigDecimal(value)
-      case JArray(elems) => Json.fromValues(elems.map(apply))
-      case JObject(fields) => Json.fromFields(fields.map { case (key, value) => (key, apply(value)) })
-      case JNull => Json.Null
-      case JNothing => Json.Null
-    }
-  }
+object CirceDecoders {
```
cb372: 💯  for getting rid of the serialization/deserialization terminology

json/src/main/scala/com/gu/contentapi/json/CirceDecoders.scala
```diff
@@ -95,6 +70,18 @@ object CirceSerialization {
...
+  implicit val editionsResponseDecoder = Decoder[EditionsResponse]
+  implicit val tagsResponseDecoder = Decoder[TagsResponse]
+  implicit val sectionsResponseDecoder = Decoder[SectionsResponse]
+  implicit val atomsResponseDecoder = Decoder[AtomsResponse]
+  implicit val packagesResponseDecoder = Decoder[PackagesResponse]
+  implicit val errorResponseDecoder = Decoder[ErrorResponse]
+  implicit val videoStatsResponseDecoder = Decoder[VideoStatsResponse]
+  implicit val atomsUsageResponseDecoder = Decoder[AtomUsageResponse]
+  implicit val removedContentResponseDecoder = Decoder[RemovedContentResponse]
+  implicit val storyPackageArticleDecoder = Decoder[com.gu.storypackage.model.v1.Article]
```
cb372: Seems weird that we need this one, especially as we already have a `Decoder[Package]` above it on line 72

json/src/test/scala/com/gu/contentapi/json/CirceRoundTripSpec.scala
```diff
@@ -143,41 +140,38 @@ class SerializationSpec extends FlatSpec with Matchers {
     checkRoundTrip[SearchResponse]("search.json")
   }
 
-  val Identical = Diff(JNothing, JNothing, JNothing)
+  def checkRoundTrip[T: Manifest : Decoder : Encoder](jsonFileName: String,
```
cb372: You don't need a Manifest anymore. That's a json4s thing.

json/src/test/scala/com/gu/contentapi/json/CirceRoundTripSpec.scala
```diff
@@ -143,41 +140,38 @@ class SerializationSpec extends FlatSpec with Matchers {
...
-    checkDiff(jsonBefore, jsonAfter)
+    jsons should not be None
+    jsons.foreach(j => checkDiff(j._1, j._2))
   }
 
-  def checkDiff(jsonBefore: JValue, jsonAfter: JValue) = {
-    val diff = Diff.diff(jsonBefore, jsonAfter)
+  val Identical = Diff(Json.Null, Json.Null, Json.Null)
+  def checkDiff(jsonBefore: Json, jsonAfter: Json) = {
+    import com.github.agourlay.cornichon.json.JsonDiff.{ diff, Diff }
```
cb372: `JsonDiff.Diff` is already imported at the top of the file

json/src/test/scala/com/gu/contentapi/json/CirceSpec.scala
```diff
@@ -1,92 +0,0 @@
...
-        }
-        ))
-      val json: JValue = JsonMethods.parse(jsonString)
-      json.extract[T]
-    }
-
-    usingCirce should be(Right(usingJson4s))
-  }
-
-  it should "be able to deserialize a Json number to a String" in {
```
cb372: This test and the next one are worth keeping. They don't do any comparison against json4s.

json/src/test/scala/com/gu/contentapi/json/utils/JsonHelpers.scala
```diff
@@ -0,0 +1,19 @@
...
+import com.google.common.io.Resources
+import io.circe.{Decoder, Json}
+import io.circe.parser._
+
+object JsonHelpers {
+  def loadJson(filename: String): String = {
+    Resources.toString(Resources.getResource(s"templates/$filename"), StandardCharsets.UTF_8)
+  }
+
+  def parseJson[T: Manifest : Decoder](rawJson: String): T = {
```
cb372: Don't need Manifest here

json/src/test/scala/com/gu/contentapi/json/JsonParserEditionsTest.scala
```diff
@@ -1,11 +1,13 @@
 package com.gu.contentapi.json
 
-import com.gu.contentapi.json.utils.JsonLoader.loadJson
+import com.gu.contentapi.client.model.v1.EditionsResponse
+import com.gu.contentapi.json.utils.JsonHelpers._
+import com.gu.contentapi.json.CirceDecoders._
 import org.scalatest.{FlatSpec, Matchers}
 
 class JsonParserEditionsTest extends FlatSpec with Matchers {
```
cb372: JsonParser has been deleted, so all these tests need to be renamed (or can be deleted as well?)

macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
```diff
@@ -135,4 +135,56 @@ object CirceScroogeMacros {
...
+          val implicitLazyEncoder = c.inferImplicitValue(lazyEncoderForType)
+          if (implicitLazyEncoder.isEmpty) c.abort(c.enclosingPosition, s"Could not find an implicit Encoder[$tpe] even after resorting to Lazy")
+
+          // Note: In theory we could use the `implicitLazyEncoder` that we just found, but... for some reason it crashes the compiler :(
+          q"_root_.scala.Predef.implicitly[_root_.shapeless.Lazy[_root_.io.circe.Encoder[$tpe]]].value"
+        }
+      }
+
+      if (isOption) q"""thrift.${name.toTermName}.map(${name.toString} -> $implicitEncoder.apply(_))"""
+      else q"""Some(${name.toString} -> $implicitEncoder.apply(thrift.${name.toTermName}))"""
```
cb372: It's safest to fully qualify everything because Scala macros are not hygienic. 

i.e. write `_root_.scala.Some` instead of `Some`.

See http://docs.scala-lang.org/overviews/quasiquotes/hygiene for more details.

json/src/test/scala/com/gu/contentapi/json/JsonParserEditionsTest.scala
tomrf1: I've deleted the lot, as they don't test anything that isn't covered by the 'round-trip' tests

json/src/test/scala/com/gu/contentapi/json/CirceSpec.scala
tomrf1: I've moved this into the 'round-trip' test file, where it also already tests CapiDateTime decoding/encoding

cb372:
👍 

>+507 −1,720

yay

tomrf1:
@cb372 thanks. I'm going to bump the version to 10 for this release, as it's very much a breaking change.

tomrf1:
@mchv I have just updated this branch to continue to remove milliseconds when encoding as json, as we discussed. I'll update the concierge branch tests to reflect this.

PR #62, opened 2016-09-19
Update of the thrift definition to support sponsorship logo dimensions

shtukas:
This change updates the thrift definition to support sponsorship logo dimensions.

It comes after 
1. https://github.com/guardian/content-api/pull/1711 , and 
2. https://github.com/guardian/content-api/pull/1714

The dimensions are published by the tag manager here: https://github.com/guardian/tags-thrift-schema/blob/master/src/main/thrift/image.thrift


LATaylor-guardian:
👍

PR #63, opened 2016-09-23
Upgrade circe version from 0.5.0-M2 to 0.5.2

shtukas:


cb372:
👍 

We've confirmed with benchmarks that this does not affect performance.

PR #65, opened 2016-10-04
Add bodyText and charCount fields

tomrf1:
This is for alexa (but possibly other apps later).
The `bodyText` does not contain any tags. The `charCount` is the number of chars in `bodyText`, and will allow us to exclude long articles.

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -998,6 +998,10 @@ struct ContentFields {
     43: optional string shortSocialShareText
 
     44: optional string socialShareText
+
+    45: optional string bodyText
+
+    46: optional i32 charCount
 }
 
 struct Reference {


models/src/main/thrift/content/v1.thrift
```diff
@@ -998,6 +998,10 @@ struct ContentFields {
     43: optional string shortSocialShareText
 
     44: optional string socialShareText
+
+    45: optional string bodyText
+
+    46: optional i32 charCount
```
tsop14: Is it worth calling it `bodyTextCharCount` or similar? There also be charCount for the body field?

tomrf1: We decided not to because the `wordCount` is actually calculated by composer based on the `bodyText`, so the naming is consistent.

LATaylor-guardian: Can we not abbreviate it? Pet Peeve lol.

mchv:
👍

PR #68, opened 2016-10-27
Add support for recipe

mchv:


models/src/main/thrift/content/v1.thrift
```diff
@@ -1630,6 +1632,8 @@ struct ItemResponse {
     25: optional contentatom.Atom interactive
 
     26: optional contentatom.Atom review
+
+    27: optional contentatom.Atom recipes
```
LATaylor-guardian: This should be: `27: optional contentatom.Atom recipe`

LATaylor-guardian:
Apart from my comment 👍

PR #70, opened 2016-11-02
add field for vidispineid

paulmr:
We would like to keep track of the vidispineId so that we can track back from a video page that was created by pluto, to that entry in pluto.

This PR is the first step. There will also be changes to composer and porter to actuall fill this field in.

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -1003,7 +1003,7 @@ struct ContentFields {
 
     46: optional i32 charCount
 
-    47: optional string internalVideoId
+    47: optional string internalVideoCode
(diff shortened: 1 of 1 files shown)


mchv:
@paulmr Is this is specific to video, why not having it only in `videoElementFields`?

paulmr:
@mchv: I am not sure. It doesn't belong to a video element, but is only applicable to an entire page that was created as a video placeholder page.

mchv:
@paulmr make sense, I will say just rename it to `internalVideoCode` just for consistency

paulmr:
@mchv good suggestion, thanks.

mchv:
👍

PR #71, opened 2016-11-02
Add Content Atom Type and ID Data to atom elements

susiecoleman:
This adds more information about content atom elements at a block level.
Adding 2 fields for content atom elements:
- Atom ID
- Atom Type

Changes discussed with @clloyd and @mchv


LATaylor-guardian:
👍

PR #77, opened 2016-11-16
Bump content-atom to 2.4.15

tomrf1:
- include legallySensitive flag for atoms (2.4.14)
- make currency value an int (2.4.15)


mchv:
Please add in description why, apart from that 👍

PR #81, opened 2016-11-28
Thrift union macros

tomrf1:
This library currently has a custom decoder and encoder for `Atom`, because `AtomData` is a thrift union type and our standard macros cannot handle unions. This is annoying because we have to update this code every time an atom is added to the model.

This change adds macros for generating decoders and encoders for scrooge's thrift union types, meaning fewer code changes in future.

### DIFF
--- a/macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
+++ b/macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
@@ -113,6 +113,7 @@ object CirceScroogeMacros {
     val tree = q"""{
       _root_.io.circe.Decoder.instance((cursor: _root_.io.circe.HCursor) => for (..${params.map(_._2)}) yield $apply(..${params.map(_._1)}))
     }"""
+    //println(showCode(tree))
     tree
   }
 
@@ -141,7 +142,7 @@ object CirceScroogeMacros {
     * Macro to produce Decoders for ThriftUnion types.
     *
     * A ThriftUnion, U, has a companion object which defines a case class extending U for each member of the union.
-    * Each case class takes a single parameter, which is a type alias for the actual ThriftStruct contained by that member.
+    * Each case class takes a single parameter, whose type is an alias for the actual type contained by that member.
     *
     * E.g. for the following thrift definition:
     * {{{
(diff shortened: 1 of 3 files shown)


macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
```diff
@@ -110,7 +113,6 @@ object CirceScroogeMacros {
     val tree = q"""{
       _root_.io.circe.Decoder.instance((cursor: _root_.io.circe.HCursor) => for (..${params.map(_._2)}) yield $apply(..${params.map(_._1)}))
     }"""
-    //println(showCode(tree))
```
mchv: could we keep this, this is useful for debugging!

mchv: I am suggesting we keep the code commented, as this is useful for debugging. the hard part of the debug session is not to add the `println` but to remember where to put it and not to forget the `showCode ` instruction 😄

tomrf1:
@philwills you suggested you could take a look at this one. Happy to talk through it IRL

macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
```diff
@@ -187,4 +281,49 @@ object CirceScroogeMacros {
...
   }
+
+  implicit def encodeThriftUnion[A <: ThriftUnion]: Encoder[A] = macro encodeThriftUnion_impl[A]
+
+  def encodeThriftUnion_impl[A: c.WeakTypeTag](c: blackbox.Context): c.Tree = {
+    import c.universe._
+
+    val A = weakTypeOf[A]
+
+    val memberClasses: Iterable[Symbol] = A.companion.members.filter { member =>
```
philwills: By using macro bundles (http://docs.scala-lang.org/overviews/macros/bundles), you can have multiple methods sharing the same context/universe, which allows you to factor out common methods like this.

macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
```diff
@@ -135,9 +138,100 @@ object CirceScroogeMacros {
...
+    *
+    */
+  implicit def decodeThriftUnion[A <: ThriftUnion]: Decoder[A] = macro decodeThriftUnion_impl[A]
+
+  def decodeThriftUnion_impl[A: c.WeakTypeTag](c: blackbox.Context): c.Tree = {
+    import c.universe._
+
+    val A = weakTypeOf[A]
+
+    val memberClasses: Iterable[Symbol] = A.companion.members.filter { member =>
```
mchv: For the record `knownDirectSubclasses` [is fixed in scala `2.12.1`](https://github.com/scala/scala/pull/5345) and the fix [will be released in scala in `2.11.9`](https://github.com/scala/scala/pull/5345)

macros/src/main/scala/com/gu/contentapi/circe/CirceScroogeMacros.scala
tomrf1: Thanks for taking a look @philwills !
Looks like using macro bundles will definitely tidy things up. I think I'll do that work in a separate PR though.

PR #83, opened 2016-12-02
Bump content-atom-model to 2.4.23

tomrf1:
New content-atom-model version with [fixed type for `year` field](https://github.com/guardian/content-atom/pull/69)

I have not introduced cross-compiling for scala 2.12 because it is incompatible with the currently-used version of circe, and upgrading to circe 0.6 will require some code changes.


LATaylor-guardian:
👍

PR #88, opened 2017-01-09
Use macro bundle

tomrf1:
### What?
Refactors the existing Thrift encoder/decoder macros to use a [macro bundle](http://docs.scala-lang.org/overviews/macros/bundles).

### Why?
The main advantage here is that it allows multiple macro implementations to share common logic, because in the bundle all the methods now share the same context and universe. See the private methods at the end of `CirceScroogeMacrosImpl` for this shared logic.

I think this makes the code a *lot* more readable (by the standards of scala macros).

With thanks to @philwills for the [original suggestion](https://github.com/guardian/content-api-models/pull/81#discussion_r91917708)


mchv:
👍

philwills:
👍

PR #90, opened 2017-01-10
Fix the format of revisionSeenByPorter

mchv:
I think it will not break anyome
Current display is wrong:
```json
"debug": {
    "lastSeenByPorterAt": "2017-01-10T12:01:58Z",
    "revisionSeenByPorter": "0145-01-01T00:00:00Z",
    "contentSource": "flexible"
},
```


tomrf1:
👍

LATaylor-guardian:
👍

PR #92, opened 2017-01-18
Use fezziwig library

tomrf1:
For compile-time generation of scrooge encoders/decoders.

Note that we do not import the library's `encodeThriftEnum` macro because our CAPI enum values are weird, and we have special logic for this.


shtukas:
LGTM

LATaylor-guardian:
I think this should be a new major version, so v11.0 as removing the macro project is a substantial change and I'm sure some projects rely on this specifically.

PR #96, opened 2017-02-23
Add categories field to Tag model

tomrf1:
`categories` - these will correspond to the [google knowledge graph entity types](https://developers.google.com/knowledge-graph/)

The [model in tag manager](https://github.com/guardian/tagmanager/blob/master/app/model/Tag.scala#L27) already has an (unused) `categories` field.

`entityIds` - these will be any associated [entities](https://github.com/guardian/content-entity/pull/6)


LATaylor-guardian:
Discussed IRL 👍

PR #100, opened 2017-03-14
Add weekend crossword type.

philmcmahon:
This is as part of a wider piece of work to get weekend crosswords being published interactively


shtukas:
:+1:

PR #101, opened 2017-03-15
Fezziwig 0.3

tomrf1:
This version supports [accumulated errors](https://github.com/guardian/fezziwig/pull/11), which content-api-models will not use, but it's good to keep it up to date.

I've rearranged some of the implicit decoder definitions as this improves compilation times. 
E.g. if `Atom` is used in multiple places and we generate the decoder for `Atom` first, then that generated implicit definition can be used rather than generating a new decoder each time.

Benchmarks are fine.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,7 +122,7 @@ lazy val json = Project(id = "content-api-models-json", base = file("json"))
   .settings(
     description := "Json parser for the Guardian's Content API models",
     libraryDependencies ++= Seq(
-      "com.gu" %% "fezziwig" % "0.2",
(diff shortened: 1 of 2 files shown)


LATaylor-guardian:
👍  (try on code first)

tomrf1:
I've pushed another commit after discovering that one of the concierge tests fails with the new version.

The problem is that when decoding to `CapiDateTime`, the json field may be a string or an object, depending on whether it's come from elasticsearch or a concierge response. The `dateTimeDecoder` was telling it to expect a string only.

I'm not sure why this was never a problem before.

PR #94, opened 2017-02-07
Circe 0.7.0

tomrf1:
and use the latest fezziwig

PR #113, opened 2017-07-10
Fezziwig 0.4

tomrf1:
This picks up 2 changes, neither of which content-api-models needs:
1. Supports [accumulated errors](https://github.com/guardian/fezziwig/pull/11)
2. Newer versions of circe and thrift/scrooge - https://github.com/guardian/fezziwig/pull/12/files

### Decoder definitions
I've rearranged some of the implicit decoder definitions as this improves compilation times.

### dateTimeDecoder
A concierge test fails without this new `dateTimeDecoder`.
This has led to the discovery that dateTime fields under `requestedBodyBlocks` are being incorrectly serialized to json as objects rather than strings. It's not clear why this is happening - I'll fix it separately.

It's also not clear why bumping the fezziwig version has broken decoding of this json back to thrift. It was somehow working before - even though the old `dateTimeDecoder` was telling it to expect a string only.

### thrift/scrooge dependencies
We have to be careful not to use the latest version, as this has caused problems for clients. Hence the `dependencyOverrides`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -117,7 +117,13 @@ lazy val scala = Project(id = "content-api-models-scala", base = file("scala"))
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % "0.9.1",
       "com.twitter" %% "scrooge-core" % "4.5.0"
-    )
+    ),
+
+    /**
+      * WARNING - upgrading the following will break clients
+      */
+    dependencyOverrides += "org.apache.thrift" % "libthrift" % "0.9.1",
+    dependencyOverrides += "com.twitter" %% "scrooge-core" % "4.5.0"
   )
 
 /**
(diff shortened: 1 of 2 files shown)


mchv:
👍

build.sbt
```diff
@@ -117,7 +117,9 @@ lazy val scala = Project(id = "content-api-models-scala", base = file("scala"))
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % "0.9.1",
       "com.twitter" %% "scrooge-core" % "4.5.0"
-    )
+    ),
+    dependencyOverrides += "org.apache.thrift" % "libthrift" % "0.9.1",
```
mchv: adding a comment to explain why to not update (or remove the override) will help ourself in the future to not break things!

tomrf1: Good point. I wish I could remember the actual reason for this!

tomrf1:
Benchmarks are fine compared with [the last time I ran them](https://github.com/guardian/content-api-models/pull/94#issuecomment-278007497) 
```
[info] # Run complete. Total time: 00:29:59
[info]
[info] Benchmark                                          Mode  Cnt        Score        Error  Units
[info] JsonDecodeBenchmark.circeDecode                    avgt  200   127841.727 ±   7173.856  ns/op
[info] JsonDecodeBenchmark.circeDecodeEachContent         avgt  200  1675967.130 ± 158535.970  ns/op
[info] JsonDecodeBenchmark.circeDecodeEachMassiveContent  avgt  200      334.921 ±      5.243  ms/op
[info] JsonDecodeBenchmark.circeEncode                    avgt  200   127772.119 ±   2246.400  ns/op
```

PR #118, opened 2017-09-15
Upgrade to scala 2.12.3

tomrf1:
Bumping the scrooge version is safe.
libthrift must not go beyond 0.9.1.
Circe 0.8.
Also switched from cornichon to diffson for json diffs in tests.


mchv:
👍 let's ship it

PR #120, opened 2017-09-19
Revert cross paths change

tomrf1:
An earlier change enabled cross publishing (2.11 and 2.12). This had the unwanted side-effect of adding the scala version to the end of the name for *all* projects published by this library:
- `content-api-models-scala_2.12` - wanted
- `content-api-models-json_2.12` - not wanted
- `content-api-models_2.12` - not wanted

This other change that I reverted here attempted to fix this, but sbt doesn't allow you to only set `crossScalaVersions` for one project in a multi-project build. So the result was that only 2.12 was published.

We've decided to accept having scala versions after all projects, as we need cross publishing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ val mavenSettings = Seq(
 
 val commonSettings = Seq(
   scalaVersion := "2.12.3",
-  crossPaths := false,
+  crossScalaVersions := Seq("2.11.8", scalaVersion.value),
   releasePublishArtifactsAction := PgpKeys.publishSigned.value,
   organization := "com.gu",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -46,7 +46,7 @@ val mavenSettings = Seq(
 
 val commonSettings = Seq(
   scalaVersion := "2.12.3",
-  crossPaths := false,
+  crossScalaVersions := Seq("2.11.8", scalaVersion.value),
```
regiskuckaertz: `2.11.11`?

tomrf1: it should be! This PR is just reverting the previous merge though

regiskuckaertz:
👍

build.sbt
regiskuckaertz: Fair enough!

PR #124, opened 2017-09-27
update to latest atom model

mchv:


mchv:
merging but please comment after if needed

PR #127, opened 2017-10-11
Make time optional in CapiDateTime decoder

tomrf1:
The recent move from joda to java time breaks the decoding of elasticsearch json in Concierge.
This is because some of these fields have only a date, and the joda time parser was less strict.

I've tested locally with concierge.

@annebyrne this was blocking your concierge version bump :(


annebyrne:
timezones

mchv:
👍

PR #128, opened 2017-10-18
Add Sponsorship field to Rich Links

katebee:
Related:

https://github.com/guardian/content-api/pull/2030

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -75,7 +75,7 @@ enum ElementType {
     COMMENT = 6,
 
     /*
-     * A rich link to guardian content - presents a nice trail outside of the contnet body
+     * A rich link to guardian content - presents a nice trail outside of the content body
      */
     RICH_LINK = 7,
 
@@ -559,6 +559,7 @@ struct RichLinkElementFields {
     3: optional string linkText
     4: optional string linkPrefix
     5: optional string role
+    6: optional Sponsorship sponsorship 
 }
 
 struct MembershipElementFields {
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -559,6 +559,7 @@ struct RichLinkElementFields {
     3: optional string linkText
     4: optional string linkPrefix
     5: optional string role
+    6: optional Sponsorship sponsorship
```
tomrf1: Is this intended to match the [model in flexible-model](https://github.com/guardian/flexible-model/blob/master/src/main/thrift/content.thrift#L403)?
Unfortunately this library doesn't depend on flexible-model, so you'll need to copy the thrift definition in here

katebee: ??? 🤔 

I believe `RichLinkElementFields` is now using this version of [Sponsorship](https://github.com/guardian/content-api-models/blob/fa31a09b0a84a8cfc754ac43e39bffae7589b29a/models/src/main/thrift/content/v1.thrift#L1076), as a transformation occurs in content-api

@LATaylor-guardian just a thought - will it be a problem that `Sponsorship` is defined below the `RichLinkElementFields`?

tomrf1: sorry I completely missed the point here :)

tomrf1: The build is currently failing, I think you're right that it's because `Sponsorship` is defined after

LATaylor-guardian: @katebee Yes, good spot. This is why it's not compiling.

tomrf1:
👍 I'll release this shortly

PR #129, opened 2017-10-19
Include podcast level info about types - serial or episodic

annebyrne:
In the iOS 11 update, we can now include podcast level information about the `type` of the podcast, whether they are serial or episodic. This will impact how episodes of podcast are shown to the user.


tsop14:
👍

PR #130, opened 2017-10-23
Add pillar fields to Content

tomrf1:
In addition to `sectionId`/`sectionName`, capi will now also return the pillar details.
Concierge will determine the pillar from the content's `sectionId`.


mchv:
👍

PR #142, opened 2018-02-13
Upgrade lib thrift and scrooge-core version to same as ophan-event-model

davidfurey:
Reduce number of binary incompatibility warnings in other projects

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,8 +121,8 @@ lazy val scala = Project(id = "content-api-models-scala", base = file("scala"))
       (scroogeUnpackDeps in Compile).value.flatMap { dir => (dir ** "*.thrift").get }
     },
     libraryDependencies ++= Seq(
-      "org.apache.thrift" % "libthrift" % "0.9.1",
-      "com.twitter" %% "scrooge-core" % "4.18.0"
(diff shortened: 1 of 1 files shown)


mchv:
@davidfurey We intentionally stayed on `0.9.1` [following a regression when we updated](https://github.com/guardian/content-api-models/pull/10). The binary compatibility warnings is one that should not be ignored I think. I added a [dependency override](https://github.com/guardian/ophan/blob/a8f3ce54a5e6658fb5412dc06ac7ca95631f907f/build.sbt#L153-L156) in `ophan` `shared-lib`, but I can see that other project add a [dependency to an upper version](https://github.com/guardian/ophan/blob/a8f3ce54a5e6658fb5412dc06ac7ca95631f907f/build.sbt#L178)!

I think we may not have `ophan` runtime errors currently because we don't enter the code path that delegates to libthrift `TField.equals` method (_struct containing one or more optional fields that it doesn't understand_)

davidfurey:
Ah, yes that sounds vaguely familiar.  Thanks for picking up on this.  Twitter are working on upgrading libthrift version so hopefully we can upgrade soon - https://github.com/twitter/finagle/issues/673

PR #143, opened 2018-04-05
Add new showAffiliateLinks field

philmcmahon:
This field will be set in composer, and used by frontend to determine whether affiliate links are displayed within an article.

Background: We're going to start automatically adding affiliate links to content in certain sections (e.g. lifeandstyle). They come from a service called skimlinks. Sometimes we'll need to manually turn them off for sensitive content, or force them to be on for content not in one of the whitelisted sections.


tomrf1:
👍

PR #145, opened 2018-06-12
Add lastModified to RemovedContentResponse

tomrf1:
Needed by crier for [new design](https://github.com/guardian/crier/pull/79).
For backwards-compatibility, this must be a new field.
It can be optionally returned by capi if clients request it

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -1629,6 +1629,12 @@ struct Pillar {
     3: required list<string> sectionIds
 }
 
+struct RemovedContent {
+    1: required string id
+
+    2: required CapiDateTime lastModified
+}
+
 /* These are Responses structures shared with the Content API */
 
 struct SearchResponse {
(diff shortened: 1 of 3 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -1828,6 +1834,8 @@ struct RemovedContentResponse {
     8: required string orderBy
 
     9: required list<string> results
+
+    10: optional list<RemovedContent> resultsWithLastModified
```
LATaylor-guardian: Maybe we can name this in way to provide more meaning? `removedContentResults` maybe?

LATaylor-guardian: I'm assuming we'd always filter this out for normal search responses?

tomrf1: This is on the `RemovedContentResponse` type, where currently `results` just gives you the removed content ids.
I'll make this new field hidden by default and add a query parameter that crier can send

PR #153, opened 2018-11-21
byline and pullquote role

AWare:
adds byline and pullquote
https://github.com/guardian/flexible-content/pull/3340


regiskuckaertz:
Argh, at some point we will have to refactor this according to usage pattern 😫

PR #154, opened 2018-12-24
Unify all the modelz

regiskuckaertz:
At the moment we have five repos in addition to this one: content-atom, content-entity, tags-thrift-schema, story-model, story-packages-model. It is hard to understand how they relate to each other without going on a hunt down the dependency tree, which is mad, but also easy to misunderstand. Adding to that:

- every update to a leaf dependency ripples up the tree in a litany of PRs, and that's just for the updates to the model

- version numbers are meaningless: it's not like all those definitions make any sense in and of themselves — except for authoring tools

- releases are time-consuming: life is too short, can we speed those up?

TODO:

- [ ] Finalise folder architecture
- [ ] Unify SBT build (oh boy, I'm not looking forward to that)
- [ ] Unify version number
- [ ] Streamline release process


ajwl:
FTR: I just got confused searching for one of those leaf dependencies `content-atom` and took a while to figure out it ends up here. So would be very happy to see this merge done

PR #160, opened 2019-03-11
Add `Tag#internalName`

regiskuckaertz:
Will be exposed to internal users only


regiskuckaertz:
[Tags internal name: Add field to the model](https://trello.com/c/c2JpOjbf/655-tags-internal-name-add-field-to-the-model)

regiskuckaertz:
The field is required on your side, so I'll make it required here too.

regiskuckaertz:
We will have to reindex all the tags though... that a bad thing, right?

PR #162, opened 2019-04-03
Upgrade to Scrooge 19.3

regiskuckaertz:


NataliaLKB:
The PR build seems to have failed on this branch

build.sbt
```diff
@@ -122,14 +122,9 @@ lazy val scala = Project(id = "content-api-models-scala", base = file("scala"))
...
     libraryDependencies ++= Seq(
-      "org.apache.thrift" % "libthrift" % "0.9.1",
-      "com.twitter" %% "scrooge-core" % "4.18.0"
-    ),
-
-    /**
-      * WARNING - upgrading the following will break clients
-      */
-    dependencyOverrides += "org.apache.thrift" % "libthrift" % "0.9.1"
+      "org.apache.thrift" % "libthrift" % "0.10.0",
```
JustinPinner: The `WARNING` (removed) makes me nervous. Are these versions backward-compatible? What about in future - do we still have to be cautious in this area?

regiskuckaertz: It was there because we thought we could not upgrade the library while risking to break every service. But it only happens when there's a negative discrepancy between producers and consumers, like when I added the `internalName` field in concierge and it broke frontend. This morning I tested frontend with the most recent version of `libthrift` and `scrooge` and it was working fine, even with Concierge relying on v0.9.1 as above. So, while we don't change the model, I can quickly update every consumer first and then tackle concierge.

regiskuckaertz:
I will only be releasing 2.12. The 2.11 is failing for some weird SBT reason, but I can't bother fixing that now. Hush hush 🏃

PR #164, opened 2019-05-23
Model non-web content

regiskuckaertz:
This is a strawman PR for thinking about non-web content in the concrete. As we discussed, the gist is that an article (`Content`) is augmented with some platform-specific context, that can be used as a filter in search queries (`?platform=web`) or as additional metadata (`?show-platforms=true`), assuming proper permissions from the consumer.

To that end, I add a `Platforms` data type that holds fields for each platform, where each is its own specific data type. There must be a least one non-empty value but thrift does not allow us to enforce that constraint. The advantage of that solution is that we keep everything in one place, unified under a single ID (the `path` of the article). However, that presumes the content and the metadata (related content, tags, story packages, etc.) will be the same across all platforms*. If we go down that path, we must make sure it is fine, otherwise we will have to add platform-specific fields to every data type inside articles, which is bad both for reasoning/writing correct code and performance.

\* this is not necessarily true for content if we can follow-up with platform-specific markup.

Alternatively, we discussed the idea of duplicating articles when content diverges, at the cost of changing the ID to `(path, platform)`. This introduces another set of tradeoffs. Notably that the content and metadata may be the same 99% of the time and that redundancy in CAPI is unnecessary, but also that it requires disambiguation logic at the boundaries of CAPI: Flexible will need to handle that redundancy, as well as Porter and Concierge. This isn't easy.

Did I summarise the design correctly?

On a side note, adding platform-specific fields opens up a can of worms, as articles in their current shape already contain some of them (`webTitle`, `newspaperEditionDate`) and it's interesting to think of how the data could be re-shaped in the future; meanwhile, a minimal amount of duplication seems harmless.

### DIFF
--- a/models/src/main/thrift/content/Platform.thrift
+++ b/models/src/main/thrift/content/Platform.thrift
@@ -15,15 +15,15 @@ struct WebPlatform {
 
   1: required string title
 
-  2: required CapiDateTime publicationDate
+  2: required CapiDateTime.CapiDateTime publicationDate
 }
 
 struct PrintPlatform {
   0: required i32 pageNumber
     
-  1: required CapiDateTime editionDate
+  1: required CapiDateTime.CapiDateTime editionDate
 }
 
 struct EditionPlatform {
-  0: required CapiDateTime editionDate
+  0: required CapiDateTime.CapiDateTime editionDate
 }
--- a/models/src/main/thrift/content/Platform.thrift
+++ b/models/src/main/thrift/content/Platform.thrift
@@ -1,6 +1,7 @@
 namespace scala com.gu.contentapi.client.model.v1
 
 include "CapiDateTime.thrift"
+include "Tag.thrift"
 
 enum Platform {
   WEB = 0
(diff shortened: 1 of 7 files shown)


models/src/main/thrift/content/Platform.thrift
```diff
@@ -0,0 +1,29 @@
...
+
+  1: required string title
+
+  2: required CapiDateTime.CapiDateTime publicationDate
+}
+
+struct PrintPlatform {
+  0: required i32 pageNumber
+    
+  1: required CapiDateTime.CapiDateTime editionDate
```
akash1810: Do we want the publication name too to differentiate `The Guardian` and `The Observer` (and `Auto Trader`)?

regiskuckaertz: I didn't think about that. Is it print only?

JustinPinner:
> that presumes the content and the metadata (related content, tags, story packages, etc.) will be the same across all platforms

I'd be surprised if this is the case very often, if at all. Maybe we could outline our first specific use case and see how it shapes up in terms of similarity with existing structures and data?

models/src/main/thrift/content/Platform.thrift
akash1810: I think so, that's certainly how we model it in Grid. @blishen will likely know for certain.

regiskuckaertz: Done by Simon.

regiskuckaertz:
Have started preparing changes in CAPI here: https://github.com/guardian/content-api/pull/2289

regiskuckaertz:
Closing, please do it eventually.

PR #165, opened 2019-06-03
Some tweaks to build on modelling non-web content

sihil:
A few key changes:
 - `platform` becomes a `publicationInstance`, i.e. a place where this content has been published
 - in the main model this is now a list as something can be published in multiple locations
 - we mimic the element structure with an enum to select the platform for a publication instance and then different structs for each type
 - we include the pub/book/book-section 'tags' in print (these might become something other than tags if better)

### DIFF
--- a/models/src/main/thrift/content/Platform.thrift
+++ b/models/src/main/thrift/content/Platform.thrift
@@ -2,28 +2,52 @@ namespace scala com.gu.contentapi.client.model.v1
 
 include "CapiDateTime.thrift"
 
-struct Platforms {
-  0: optional WebPlatform web;
-
-  1: optional PrintPlatform print;
-
-  2: optional EditionPlatform edition;
+enum Platform {
+  WEB = 0
+  PRINT = 1
+  EDITION = 2
 }
 
-struct WebPlatform {
-  0: required string url
+struct PublicationInstance {
+  // let's have a platform enum to avoid ambiguity of this struct (only fields
+  // in the selected platform are relevant)
+  0: required Platform platform
+
+  1: optional WebFields webFields
+
+  2: optional PrintFields printFields
+
+  3: optional EditionFields editionFields
+}
+
+struct WebFields {
+  0: required string url // isn't this determined by the path???
 
   1: required string title
-
-  2: required CapiDateTime.CapiDateTime publicationDate
 }
 
-struct PrintPlatform {
-  0: required i32 pageNumber
-    
-  1: required CapiDateTime.CapiDateTime editionDate
+struct PrintFields {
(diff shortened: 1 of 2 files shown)


models/src/main/thrift/content/Platform.thrift
```diff
@@ -2,28 +2,52 @@ namespace scala com.gu.contentapi.client.model.v1
...
+  // in the selected platform are relevant)
+  0: required Platform platform
+
+  1: optional WebFields webFields
 
-  1: optional PrintPlatform print;
+  2: optional PrintFields printFields
 
-  2: optional EditionPlatform edition;
+  3: optional EditionFields editionFields
```
philmcmahon: should we use the union type here since if something is in WEB platform it will never have PrintFields?

philmcmahon: (though caveat: unions can be a pain to work with in scala as we have to pattern match everywhere)

LATaylor-guardian: Historically we've tended to avoid them for this reason as they make the SerDe code more complex.

models/src/main/thrift/content/v1.thrift
```diff
@@ -1553,7 +1553,10 @@ struct Content {
 
     25: optional string pillarName
 
-    26: optional Platform.Platforms platforms
+    /*
+     * Where has this piece of content been published?
+     */
+    26: optional list<Platform.PublicationInstance> publicationInstances
```
regiskuckaertz: My fear is that this will make querying slow and introduce much complexity in the code. The way I see it is that we have three options:

- **One item per article in the index**, with all platform-specific data distilled into the `PublicationInstance`s
  - pros: one ID to rule them all, backwards-compat is just adding a filter in Concierge, no downstream change to clients
  - cons: implies the content is the same everywhere; not necessarily bad, esp. if we follow-up with platform-specific markup.

- **One item per platform per article in the index**, i.e. each item in the index exists for a single platform
  - pros: more flexibility with schema evolution in the future instead of shoving everything into `PublicationInstance`s, the content of each item (metadata and all) can be fully customised per platform
  - cons: the ID becomes a tuple `(path, platform)` that needs (de-)normalisation, I'm not sure we need that level of flexibility

- **One item per list of platforms in the index**, i.e. one item can be use for several platforms
  - pros: less redundancy than option 2 but same flexibility with the content
  - cons: need for a new ID, risk of introducing inconsistency (duplicates) pushes the complexity down into CAPI

With option 2 and 3, querying by ID will need additional filters and denormalising. Just flagging this out, if we can live with option 1 at the beginning it seems to be less effort.

PR #168, opened 2019-08-14
Unique publication instance

regiskuckaertz:
Just want to record the fruit of our session last week with @blishen @akash1810 @jonathonherbert  and @JustinPinner 

We adopt a "forking" model:

- a piece of content can only belong to one platform: this is because changes in the article made for that platform (which happens more often that not) cannot be reflected in the other instances. This is both true when a fork is for another platform or for the same platform as a reissue.

- in reflection, adding platform-specific annotations and content can quickly make the UI very complex and hard to understand and pushing that complexity "upwards" seems like a more realistic approach at the beginning: it is far easier to add a "fork/switch" UI.

- at the moment we don't track parent/child/sibling relationships in CAPI, but this can easily be added if necessary.

### DIFF
--- a/models/src/main/thrift/content/Platform.thrift
+++ b/models/src/main/thrift/content/Platform.thrift
@@ -3,22 +3,12 @@ namespace scala com.gu.contentapi.client.model.v1
 include "CapiDateTime.thrift"
 include "Tag.thrift"
 
-enum Platform {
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
I won't have time to work on this, feel free to pick it back up.

PR #170, opened 2019-09-16
Remove stories

regiskuckaertz:
Stories have been deleted from the scala client (see https://github.com/guardian/content-api-scala-client/pull/289) and CAPI (see https://github.com/guardian/content-api/pull/2304) so the models can also be cleaned out - when any clients have been updated.


regiskuckaertz:
[Remove stories in models](https://trello.com/c/0SaDoUmP/811-remove-stories-in-models)

PR #172, opened 2019-10-04
Use 0.11 for scala 2.11

regiskuckaertz:
Re-enable publishing for Scala 2.11 as I think some projects might still require it.


build.sbt
```diff
@@ -53,7 +53,24 @@ val commonSettings = Seq(
...
+      "io.circe" %% "circe-optics" % circeVersion,
+    )
+  case _ => 
+    val circeVersion = "0.12.0"
+    Seq(
+      "io.circe" %% "circe-core" % circeVersion,
+      "io.circe" %% "circe-generic" % circeVersion,
+      "io.circe" %% "circe-parser" % circeVersion,
+      "io.circe" %% "circe-optics" % circeVersion,
+    )
```
JustinPinner: These are exact duplicates - can we eliminate that?

PR #179, opened 2020-06-04
Support typescript

alexduf:
## What does this change?

- Upgrade the release pipeline
- Add support for typescript
- Update the readme

~This PR currently doesn't compile as it relies on the [yet-to-be-released story packages version](https://github.com/guardian/story-packages-model/pull/13).~ => released!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,21 +3,27 @@
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.gu/content-api-models_2.12/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.gu/content-api-models_2.12)
 [![Build Status](https://travis-ci.org/guardian/content-api-models.svg?branch=master)](https://travis-ci.org/guardian/content-api-models)
 
-To release:
+To release, in the SBT repl:
 
-```
-$ sbt 'release cross'
+```sbtshell
+release cross // will release the scala / thrift projects
+project typescript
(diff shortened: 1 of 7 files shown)


alexduf:
- The last travis build passes (although github doesn't update for some reason)
- I've fixed a gap I had in the thrift compiler https://github.com/guardian/scrooge-extras/pull/8
- Both the scala and the typescript compiles locally now

As a not I've found [a bug](https://github.com/npm/cli/issues/1392) in NPM which means when publishing this library I'll have to go back to using a three part version number: `15.10.0` rather than `15.10` is that okay?

I believe this can now be reviewed

JustinPinner:
Hi @alexduf, this looks good to me. Would you add a comment to the README about the three part version number problem? E.g. what are the implications if someone got that wrong for example? Would the release just fail or might we accidentally replace an existing released version (I don't _think_ that's possible, but if it's a bug then weird things might happen)?

I'll go ahead and approve, and will add a card for us to check the implications for the CAPI scala client. Would you be available to lend a hand with that if needed?

alexduf:
>Would you be available to lend a hand with that if needed?

Yes!

>E.g. what are the implications if someone got that wrong for example?

The maven release would be fine but the npm release would fail

I think I need to add a check in the sbt plugin before trying to release so I ensure the version number is correct

I'll add something to the readme and merge. Thanks for the review!

PR #183, opened 2020-11-18
Element tracking field is promoted from embed element to all block elements

tonytw1:
Optional field which will render of the element is deemed to potentially contain tracking embeds

Screenshot 2020-11-18 at 10 40 36

Optional tracking field shown appearing on an embed and text element.

## What does this change?

Promotes the track field for general use on all applicable block elements.

This field has never used in it's initial position under the embed element. Promoting it is considered safe.

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -759,6 +759,11 @@ struct BlockElement {
 
     20: optional ContentAtomElementFields contentAtomTypeData
 
+    /*
+     * Optional field which may appear on a block element with a html field containing third party embed code.
+     * Likely to appear on embed and other rich media type elements.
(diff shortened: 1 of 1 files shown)


tonytw1:
The text element in the above screen shot is an interesting example.
The tracking field is optional; it probably doesn't need to be rendered for text elements which we know do not contain embed code.

models/src/main/thrift/content/v1.thrift
```diff
@@ -760,6 +759,8 @@ struct BlockElement {
 
     20: optional ContentAtomElementFields contentAtomTypeData
 
+    21: optional EmbedTracking tracking
```
Mullefa, line 762: Think it would be helpful to include a comment as to why this field is optional.

Mullefa (approved):

PR #184, opened 2020-11-27
Element role additions

tonytw1:
## What does this change?

Adds the 'role' field to Block Element types which did not already have it; notably embed and atom.

As per: https://trello.com/c/cuA5lpWH/1290-embed-role-not-modelled-in-capi-block-element-view

## How to test

Adding a embed element in Composer should result in the role field appearing in the CAPI Block Element.

This is a field addition. It should be safe.


frj (approved):
Looks good to me.

PR #185, opened 2020-12-11
The RetrievableContent event adds a content type convenience field

tonytw1:
To help consumers decide if they are interested in this update before incurring the cost of fetching it.


JustinPinner (approved):
Sure, why not? 👍

PR #187, opened 2021-01-06
New event to represent a successful content decache

tonytw1:
## What does this change?

Introduces an event to represent Fastly CDN decaches.

Includes some basic meta data which the consume may find useful.

Intended as a lightweight substitute of the crier Event which can be iterated on to better suit the decached content use case.

### DIFF
--- a/models/src/main/thrift/events/fastly/event.thrift
+++ b/models/src/main/thrift/events/fastly/event.thrift
@@ -1,0 +1,28 @@
+namespace scala com.gu.fastly.model.event.v1
+#@namespace typescript _at_guardian.content_api_models.fastly.event.v1
+
+include "content/v1.thrift"
+
+enum EventType {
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/events/fastly/event.thrift
```diff
@@ -0,0 +1,28 @@
...
+    * Convenience field to help consumers filter events
+    */
+    3: optional v1.ContentType contentType
+
+   /*
+    * If the content's URL has evolved over time, we include
+    * the aliasPaths so that e.g. de-caching can be comprehensively
+    * triggered when the content is updated
+    */
+    4: optional list<string> aliasPaths
```
JustinPinner, line 27: 👏👏👏

JustinPinner (approved):
Looks grand to me 👍

PR #188, opened 2021-01-15
Deleted Content payload

tonytw1:
## What does this change?

Provides a new type of content update payload to document the the optional alias paths of a piece of deleted content.

A content deleted update may start to provided this payload; currently content deleted events have no payload set.

### DIFF
--- a/models/src/main/thrift/events/crier/event.thrift
+++ b/models/src/main/thrift/events/crier/event.thrift
@@ -59,6 +59,20 @@ struct RetrievableContent {
     6: optional list<string> aliasPaths
 }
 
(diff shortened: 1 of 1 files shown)


JustinPinner (approved):
LGTM. Might be good to get another opinion too though (as I had a hand in making this change). Otherwise 👍

tonytw1:
> LGTM. Might be good to get another opinion too though (as I had a hand in making this change). Otherwise 👍

Yeah - slight role reversal. I'm endorsing your original commit be doing this PR so I think it'll stand up in court =)

PR #189, opened 2021-01-15
Adding source fields to PullquoteElementFields and EmbedElementFields

frj:
All elements that can contain embedded content should have a source field
indicating the origins of the embedded content.

Most elements in question had this field apart from  PullquoteElementFields and EmbedElementFields

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -395,6 +395,8 @@ struct PullquoteElementFields {
     2: optional string attribution
 
     3: optional string role
+
+    4: optional string source
 }
 
 struct TweetElementFields {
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -395,6 +395,8 @@ struct PullquoteElementFields {
     2: optional string attribution
 
     3: optional string role
+
+    4: optional string source
```
tonytw1, line 399: Would source fit better if it was on the EmbedTracking object rather than the Element?

tonytw1 (approved):

frj:
[Consent overlays on front ends - Proof of concept](https://trello.com/c/K5hTLga3/1328-consent-overlays-on-front-ends-proof-of-concept)

PR #190, opened 2021-01-18
Update sbt sonatype plugin to latest

tonytw1:
## What does this change?

Latest sonatype plug in.
Specifically the 3.9 series includes:

```
Add sonatypeTimeoutMillis to wait longer time until close/promote stage completes. The default is 60 minutes.
```

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "15.9.11-SNAPSHOT"
+version in ThisBuild := "15.9.11"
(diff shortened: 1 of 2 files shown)


JustinPinner (approved):
LGTM 👍

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "15.9.11-SNAPSHOT"
```
frj, line 1: Should this be in this PR?

JustinPinner: I _think_ that update happens direct to master when the release is done?

tonytw1: Yeah - that's a mistake. Dirty workspace.
 Reverting.

PR #191, opened 2021-01-19
Adding height and width to elements that can contain embedded content

frj:
This change is part of the work to render a privacy 'overlay' for embedded content that requires the user to consent/click the overlay before the embed isloaded. 

To do this we need to have a correct size to render the overlay.

The height and width fields already exist in a lot of the element type data types, this change ensures all the element types that can contain embedded content have a height and width field.

This is an addition of an optional field so is low risk


tonytw1 (approved):
This looks as expected; all new fields are optional so it's safe.

PR #192, opened 2021-01-20
Adding sourceDomain to the '*ElementFields'

frj:
Adding sourceDomain to the 'ElementFields' for elements that have embedded content

This will identify where the embedded content comes from and will be
useful where a human readable 'source' is not available


JustinPinner (approved):
Looks sensible to me 👍

PR #195, opened 2021-02-12
Content decached message clean up.

tonytw1:
Rename content id to content path; as it is a path which Fastly has decached.

Retire the aliasPath field as content items which multiple paths will issue multiple decaches; consumers should react to each seperate decached path.

### DIFF
--- a/models/src/main/thrift/events/fastly/event.thrift
+++ b/models/src/main/thrift/events/fastly/event.thrift
@@ -9,7 +9,7 @@ enum EventType {
 }
 
 struct ContentDecachedEvent {
-    1: required string contentId
+    1: required string contentPath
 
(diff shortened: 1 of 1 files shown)


models/src/main/thrift/events/fastly/event.thrift
```diff
@@ -9,7 +9,7 @@ enum EventType {
 }
 
 struct ContentDecachedEvent {
-    1: required string contentId
+    1: required string contentPath
```
tonytw1, line 12: This rename allowed as Facebook is the only consumer and that's my job to update.

Mullefa (approved):

JustinPinner (approved):
Nice 👍

PR #196, opened 2021-02-16
Platform types and embed reach model items

tonytw1:
## What does this change?

Introduce the Embed Reach struct which can be used to describe embed element reach / visibility on different platforms.

Introduce the PlatformTypes enum as a way to describe platforms; this is currently a subset of those used in Ohpan.


frj (approved):
👍

PR #197, opened 2021-02-24
Add the ceasedToBeCanonicalAt datetime to evolved URL aliasPaths

JustinPinner:
## What does this change?
Ophan require visibility of the date at which an alias ceased to reflect the canonical path for the content. This PR restructures the thrift model by removing the previous `string` representation and applies a new `AliasPath` type.

## How to test
Ensure that aliasPaths are properly represented in Concierge responses, capi scala client models, crier events, pubflow monitoring events and fastly decache events.

- [x] Tested on LOCAL environment (as much as possible)
- [ ] Tested on CODE environment

The correct data appears in the correct places in the correct format? Success!

We're now changing more of the existing model but the feature isn't widely adopted yet, and because use of these are opt-in we don't imaging seeing any unexpected failures.

## Images
N/A

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -1518,6 +1518,13 @@ struct Debug {
     4: optional string originatingSystem
 }
 
+struct AliasPath {
+
+    1: required string path
+
+    2: required CapiDateTime ceasedToBeCanonicalAt
+}
+
 struct Content {
 
     /*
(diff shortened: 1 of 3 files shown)


frj (approved):
Looks good!

models/src/main/thrift/content/v1.thrift
```diff
@@ -1518,6 +1518,13 @@ struct Debug {
     4: optional string originatingSystem
 }
 
+struct AliasPath {
+
+    1: required string path
+
+    2: required CapiDateTime ceasedToBeCanonicalAt
```
tonytw1, line 1525: CapiDateTime is interesting; is there nothing more global available?
Looks like it's existing though so happy probably consistant with what other fields do.

tonytw1 (approved):
Looks legit.
Retires fields which were optional which is safe.

models/src/main/thrift/content/v1.thrift
JustinPinner: Yes, I just took a lead from elsewhere. I think it's to provide iso8601 formatting as standard.

PR #36, opened 2016-07-01
Add isHosted field to Content model.

LATaylor-guardian:

PR #198, opened 2021-03-19
Add code element type fields

Fweddi:
## What does this change?

See [#36](https://github.com/guardian/flexible-model/pull/36), [#2586](https://github.com/guardian/content-api/pull/2586) and [#2596](https://github.com/guardian/content-api/pull/2596).

Adds the 'code element type' to the capi model, along with html and language type fields.

This is ultimately to allow blocks of code to work in our developer blogs (previously these have been handled by html injection rather than content modelling).

## How to test

I think we need to deploy a full pipeline from flex to concierge, including the models, and test the behaviour. But I need a second opinion on this.

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -778,6 +778,14 @@ struct ContentAtomElementFields {
 
 }
 
+struct CodeElementFields {
+
+  1: required string html
+
+  2: required string language
+
+}
+
 struct BlockElement {
 
     1: required ElementType type
@@ -826,6 +834,8 @@ struct BlockElement {
      * Does not appear on text block elements as the html field of a text element does not contain embed code.
      */
     21: optional EmbedTracking tracking
+
+    22: optional CodeElementFields codeTypeData
 
 }
 


Fweddi:
I worry there is a conflict between the work here and the work on [Concierge](https://github.com/guardian/content-api/blob/70d8220983e6f6db4b8fbc663874f5c1a5d19fcd/concierge/src/main/scala/com.gu.contentapi.concierge/processors/elementsgenerator/CodeElementsGenerator.scala). There is a confusing difference between `typeData` (defined on the Asset struct, itself defined on the Element Struct) and `codeTypeData` (defined on the BlockElement struct). The code element type does not have any assets. Does this mean we don't need a `typeData` field, and potentially don't need to add a CodeElementGenerator at all?

mchv:
@Fweddi I think you are mixing two things:
- `show-blocks` which will show code and need changes from this PR
- `show-elements` which will triggers `CodeElementGenerator` and related code

Fweddi:
> @Fweddi I think you are mixing two things:
> 
> * `show-blocks` which will show code and need changes from this PR
> * `show-elements` which will triggers `CodeElementGenerator` and related code. We may not need `codeTypeData` in elements, and I will sip without first.

Thanks Mariot - this cleared up things a lot!

PR #199, opened 2021-04-13
Update thrift to 0.13.0

JustinPinner:
## What does this change?
Snyk was starting to grumble about our thrift version, so we've updated it to `0.13.0` (this is the most compatible with our other dependencies at the moment).

## How to test
The plan would probably be to incorporate these models into e.g. [content-api-scala-client](https://github.com/guardian/content-api-scala-client) and run some tests with that. 

Nothing falls apart or has tantrums? Success!

It might not work as expected, in which case we wouldn't release it as-is. We ought to be able to have locally built test releases that we can use to figure that out before it becomes a problem though.

## Images
happy

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,30 @@ project typescript
 releaseNpm <version> // you have to specify the version again i.e releaseNpm 1.0.0
 ```
 
+### Releasing SNAPSHOT or release candidate versions
+
+It's also possible to release a snapshot build to Sonatype's snapshot repo with no promotion to Maven Central. This can be useful for trialling a test or upgraded dependency internally.
+
+To do this, start sbt with a RELEASE_TYPE variable;
+
+`sbt -DRELEASE_TYPE=snapshot`
+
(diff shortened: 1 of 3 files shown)


JustinPinner:
I've tested this by:

* Running +publishLocal (15.10.3-SNAPSHOT)
* Update Concierge's `build.sbt` locally to use models `15.10.3-SNAPSHOT`
* Compiled then ran tests in `ContentSpec.scala` in Concierge

Found three failed tests that are due to local execution I believe, e.g. where the test domain or port doesn't match what's expected. But 35 other tests passed.

Suggest releasing this models build and then consume them in Concierge, CAPI scala client etc.

frj (approved):
Looks good, assuming sonatype will accept the snapshot version 👍

JustinPinner:
Are we happy bundling up the `build.sbt` changes to support releasing snapshots and release candidates with the far more trivial bump of thrift in one PR? If not I can separate them.

frj (approved):
Looks good/

PR #202, opened 2021-05-11
Import updated sbt-scrooge-typescript plugin 1.4.0

JustinPinner:
## What does this change?
There's a newer version of the sbt-scrooge-typescript sbt plugin available (as a result of https://github.com/guardian/scrooge-extras/pull/18) which removes the last dependency on thrift `0.12.0`. We hope.

## How to test
This change should be invisible unless the updates in the plugin have unexpected side effects. I'll aim to release this as a release candidate build and test with the CAPI scala client. Apps Rendering may have to do the same to explore whether they have issues with Int64 conversions.

If nobody finds anything wrong with the release candidate build we can proceed with a full release.

It's possible that clients may run into Int64 parsing problems but we ought to have enough test coverage to discover that quite quickly.

## Images
N/A

## Accessibility
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,13 @@
 # Content API Thrift models
 
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.gu/content-api-models_2.12/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.gu/content-api-models_2.12)
 [![Build Status](https://travis-ci.org/guardian/content-api-models.svg?branch=master)](https://travis-ci.org/guardian/content-api-models)
+
+## Version Info
+
+### 17.0.0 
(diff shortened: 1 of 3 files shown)


JustinPinner:
We'll release this as a major version increment (17.0.0) along with an additional note in the README to explain why, so we can set this back to a Ready for Review state I think.

Fweddi:
LGTM - when you say 

> removes the last dependency on thrift 0.12.0

Does this mean we are updating our thrift dependency (in `sbt-scrooge-typescript`) to 0.13.0, to keep it line with recent changes (#199?)

Fweddi (approved):

JustinPinner:
Yes, there was a reference to 0.12.0 left in the NPMLibraries file which has now gone, so we should be at 0.13.0 everywhere now. See [this PR](https://github.com/guardian/scrooge-extras/pull/18)

davidfurey:
Looks reasonable to me, but I was just wondering why you've chosen to drop the Maven central version badge?

JustinPinner:
The badge seems to have got stuck at v 14.1 and has been for ages. There didn't seem to be an alternative available for the releases we're making. If you happen to know of a way to make it work again, I'll happily put it back in.

JustinPinner:
Badge restored!

PR #203, opened 2021-05-26
Rk/fp/caption field on embeds

rowannekabalan:
## What does this change?
Allows embed type elements to have a caption, and allows CAPI to export these captions to the outside world through Concierge and CSC

-New build version is 17.1.0-SNAPSHOT
-Next build version has been updated to 17.1.1-SNAPSHOT

## How to test
Has been successfully imported into local CAPI using local publish 
Further testing can be done after importing the latest build into Concierge and CSC


JustinPinner (approved):
Looks good to me, based on https://github.com/guardian/flexible-model/pull/38 👍

PR #204, opened 2021-06-22
Updating readMe to make release process clearer

rowannekabalan:
## What does this change?
Added further clarification on the release process to README.md to make it clearer that releases should be made from main not from branches.


tonytw1 (approved):

JustinPinner (approved):
👍

PR #206, opened 2021-09-17
Mark content decached events with a published timestamp

tonytw1:
## What does this change?

Allows consumers to make informed decisions about older messages.
ie. This happened a week ago; I don't need to be retry it so aggressively or I can drop it.

- Optional for backwards compatiblity

- Choose not to use the fill CapiDateTime; these events aren't really about CAPI; it's what the crier package does; avoids cross package linkage.

- published field name could have been dateTime (crier) or firstPublished etc. published seems the simplest descriptive verb.
 

## Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#screen-readers)
- [ ] [Navigable with keyboard](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#keyboard-navigation)
- [ ] [Colour contrast passed](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/source/blob/main/docs/06-accessibility.md#use-of-colour)

### DIFF
--- a/models/src/main/thrift/events/fastly/event.thrift
+++ b/models/src/main/thrift/events/fastly/event.thrift
@@ -30,5 +30,5 @@ struct ContentDecachedEvent {
     * When was this decache event originally published
     * Date times are represented as i64 - epoch millis
     */
-    5: optional i64 published
+    5: optional i64 eventPublished
 }
--- a/models/src/main/thrift/events/fastly/event.thrift
+++ b/models/src/main/thrift/events/fastly/event.thrift
@@ -27,7 +27,7 @@ struct ContentDecachedEvent {
     */
 
    /*
-    * When was this decache event published
+    * When was this decache event originally published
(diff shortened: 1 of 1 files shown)


davidfurey:
I'm concerned that `published` could be confused with the time the article was published.  There seems to be a convention of prefixing `event` or `content` to field names to avoid ambiguity.  Perhaps `eventPublished` or `eventTimestamp` would be clearer?

tonytw1:
> I'm concerned that `published` could be confused with the time the article was published. There seems to be a convention of prefixing `event` or `content` to field names to avoid ambiguity. Perhaps `eventPublished` or `eventTimestamp` would be clearer?

eventPublished is fine.

davidfurey (approved):

PR #225, opened 2023-07-13
Add category to the Asset field

sophie-macmillan:
## What does this change?
Adds an optional Category to the `AssetFields`, this is to support the category added to the media atom so that we can show it in DCR in the video 'pill'.

The category is already supported in the [media-atom-model](https://github.com/guardian/content-atom/blob/master/thrift/src/main/thrift/atoms/media.thrift#L99)

## Other related PRs
- **content-api-models (this one)**
- content-api-scala-client: https://github.com/guardian/content-api-scala-client/pull/389
- frontend: https://github.com/guardian/frontend/pull/26316
- frontend: https://github.com/guardian/frontend/pull/26317
- DCR: https://github.com/guardian/dotcom-rendering/pull/8277

## Testing
This has all been tested locally with snapshots

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -377,6 +377,8 @@ struct AssetFields {
   68: optional bool safeEmbedCode
 
   69: optional bool isMandatory
+
+  70: optional string category
 }
 
 struct Asset {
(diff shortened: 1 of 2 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -377,6 +377,8 @@ struct AssetFields {
   68: optional bool safeEmbedCode
 
   69: optional bool isMandatory
+
+  70: optional string category
```
Divs-B, line 381: Thanks for raising this @sophie-macmillan, 
small query for information only - what values as string we can expect in field `category`?

PR #230, opened 2023-11-23
Bump content atom version

rowannekabalan:
Bumping content atom to 4.0.0 to deprecate recipe atom


tonytw1 (approved):
And is then likely to make a release of this library.
This looks fine.

PR #233, opened 2024-01-31
Add list element

simonbyford:
## What does this change?

Adds a new element type `List` which will be used to provide more structure around existing formats such as Key Takeaways.

This block element is novel in that it also contains block element, creating a recursive data structure. It is not currently anticipated that a `List` element will contain `List` elements, although there is no easy way to prevent this in the thrift definition.

The recursion did cause problems with Fezziwig's auto serialisation, which we have worked around by explicitly caching the serialisers for BlockElement. It is possible that we will change our approach here following work on Fezziwig itself by Emily Bourke.

## How to test

Run tests, they should pass.

publishLocal and try building/testing content-api-scala-client and content-api with this library.
- [x] concierge
- [x] content-api-scala-client

### DIFF
--- a/json/src/main/scala/com/gu/contentapi/json/CirceDecoders.scala
+++ b/json/src/main/scala/com/gu/contentapi/json/CirceDecoders.scala
@@ -98,7 +98,7 @@ object CirceDecoders {
   implicit val assetDecoder = Decoder[Asset]
   implicit val elementDecoder = Decoder[Element]
   implicit val referenceDecoder = Decoder[Reference]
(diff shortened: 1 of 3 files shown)


simonbyford (commented):
Looks good to me. Will need approval from someone else in the team!

davidfurey:
This change has been tested locally with data flowing from composer through to concierge

emdash-ie (approved):

davidfurey:
This has been tested locally and now in CODE with the preview release

davidfurey:
See also https://stackoverflow.com/questions/34399288/caching-implicit-resolution

PR #237, opened 2024-02-22
Recipe schema org thrift

frederickobrien:
The first step in finally getting recipe schema onto the Guardian website and reaping the SEO benefits. This PR adds a Thrift definition for Schema.org recipes.

Companions downstream: 

- https://github.com/guardian/content-api-scala-client/pull/408
- https://github.com/guardian/content-api/pull/2858
- https://github.com/guardian/facia-scala-client/pull/310
- https://github.com/guardian/frontend/pull/26931

For more information on the fields included see https://schema.org/Recipe.

Preview deployed via https://github.com/guardian/gha-scala-library-release-workflow/blob/main/docs/making-a-release.md#choose-release-type

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -1,7 +1,7 @@
 include "story_package_article.thrift"
 include "contentatom.thrift"
 include "entity.thrift"
-include "schemaorg.thrift"
+include "schema_org.thrift"
 
 namespace scala com.gu.contentapi.client.model.v1
 #@namespace typescript _at_guardian.content_api_models.v1
@@ -1684,10 +1684,6 @@ struct ContentChannel {
 
 }
 
-struct SchemaOrg {
-    1: optional list<schemaorg.SchemaRecipe> recipe
-}
-
 struct Content {
 
     /*
@@ -1832,7 +1828,7 @@ struct Content {
     /*
     * Schema.org compatible data
      */
-    29: optional SchemaOrg schemaOrg
+    29: optional schema_org.SchemaOrg schemaOrg
 }
 
 struct NetworkFront {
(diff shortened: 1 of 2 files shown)


models/src/main/thrift/content/v1.thrift
```diff
@@ -1683,6 +1684,10 @@ struct ContentChannel {
 
 }
 
+struct SchemaOrg {
+    1: optional list<schemaorg.SchemaRecipe> recipe
+}
+
```
emdash-ie, line 1690: Is this better here or in `schemaorg.thrift`? I’m not sure, would appreciate your thoughts.

On the one hand, assuming we add more schemas here in future, if we ever want to reuse this definition in other places it’s probably nicer to have it in `schemaorg.thrift`.

On the other, maybe we don’t want it to be reused anywhere, and it makes more sense to stay here.

models/src/main/thrift/content/schemaorg.thrift
emdash-ie, line 1: I find the file name `schemaorg.thrift` slightly hard to read. It’s a pretty minor preference, but what would you say to `schema_org.thrift`? (My ideal would be `schema-org.thrift`, but that’s an entirely new style, whereas we already have `contentatom.thrift` and `story_package_article.thrift`.)

emdash-ie (approved):
This looks good to me, but I’d appreciate delaying merging to discuss the (minor) points in my comments, if that’s ok?

models/src/main/thrift/content/schemaorg.thrift
frederickobrien: I'm partial to a dash myself but feel consistency is probably the best approach here. Have changed with [`45423ba` (#237)](https://github.com/guardian/content-api-models/pull/237/commits/45423ba8ff1935aff22fd207ce460b422033aed1)

emdash-ie: (I like dashes mainly because they don’t require me to press shift to type them)

PR #239, opened 2024-02-23
Update fezziwig to v2.0.0

emdash-ie:
# What does this change?

This PR updates the version of fezziwig to 2.0.0, which includes the changes from [fezziwig PR#48](https://github.com/guardian/fezziwig/pull/48), and updates the CirceEncoders and CirceDecoders as required.

Unfortunately this requires a lot of boilerplate semiauto derivation! But fortunately, it’s a little less mysterious: we no longer need fully manual encoders for Map-containing types, for example. Compile times are also somewhat improved, as described in that PR.

### DIFF
--- a/json/src/test/scala-2.12/com/gu/contentapi/json/CirceRoundTripSpec.scala
+++ b/json/src/test/scala-2.12/com/gu/contentapi/json/CirceRoundTripSpec.scala
@@ -7,7 +7,8 @@ import io.circe.syntax._
 import io.circe.parser._
 import io.circe.optics.JsonPath._
 import org.scalatest.{FlatSpec, Matchers}
-import com.gu.fezziwig.CirceScroogeMacros.{decodeThriftEnum, decodeThriftStruct, decodeThriftUnion, encodeThriftStruct, encodeThriftUnion}
+import com.gu.fezziwig.CirceScroogeMacros.{decodeThriftEnum}
+import com.gu.fezziwig.CirceScroogeWhiteboxMacros.{thriftStructLabelledGeneric, thriftUnionLabelledGeneric}
 import com.gu.contentapi.json.CirceEncoders._
 import com.gu.contentapi.json.CirceDecoders._
 import cats.syntax.either._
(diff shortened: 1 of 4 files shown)


json/src/test/scala-2.12/com/gu/contentapi/json/CirceRoundTripSpec.scala
```diff
@@ -166,20 +167,29 @@ class CirceRoundTripSpec extends FlatSpec with Matchers {
...
-      decoded <- transformedBefore.as[T].toOption
+  def checkRoundTrip[T : Decoder : Encoder](
+    jsonFileName: String,
+    transformBeforeDecode: Json => Json = identity,
+    transformAfterEncode: Json => Json = identity
+  ) = {
+    val jsonBefore: Json = parse(loadJson(jsonFileName)).toOption.get
+    val transformedBefore: Json = jsonBefore.hcursor.downField("response").success.map(
+      c => transformBeforeDecode(c.value)
+    ).get
```
paulmr, line 178: do we definitely want these two lines to throw an exception? I know it's only a test but at the same time we are going through effort later to provide a nicer failure if the second part of the round trip fails.

paulmr (approved):
Great stuff! I won't pretend I have gone through all of the boilerplate code on this one, but it looks good in principle, and so I am happy to trust the compiler + your extensive knowledge + the tests! 👍

json/src/test/scala-2.12/com/gu/contentapi/json/CirceRoundTripSpec.scala
emdash-ie: Fair enough, it was mostly laziness! I’ll update

emdash-ie: Have added [commit 9c266ba](https://github.com/guardian/content-api-models/pull/239/commits/9c266ba8a0c23a7813f8711e69c503a807ce5f46) to handle these error cases more nicely

PR #240, opened 2024-02-29
Add quick-cryptic crossword type

andrew-nowak:
Adds the upcoming Quick-Cryptic crossword type to the CAPI models. Required to have the crossword type correctly parsed from ES and returned in the Concierge response.

Tested as a local snapshot and confirmed working end-to-end through local Crosswordv2/Composer/CAPI

Before:

```
...
"crossword": {
  "type": "enum-unknown-crossword-type-1",
...
```

After:

```
...
"crossword": {
  "type": "quick-cryptic",
...
```

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -193,7 +193,9 @@ enum CrosswordType {
 
     DIAN_QUIPTIC_CROSSWORD = 6,
 
-    WEEKEND = 7
+    WEEKEND = 7,
+
+    QUICK_CRYPTIC = 8
 }
 
 enum Office {


jonathonherbert (approved):
Seems reasonable. Who the heck is `DIAN` 😀 

Point of interest – what maps `QUICK_CRYPTIC` to `quick-cryptic`? It doesn't look like it's ~scrooge~ [fezziwig.](https://github.com/guardian/fezziwig/blob/573ac54b90fd3c037aa1aa581352c3c53699a63a/src/test/scala/com/gu/fezziwig/FezziwigTests.scala#L30)

andrew-nowak:
yep, both those questions had me scratching my head, but since I was getting the answer I wanted I didn't question it too much -- now have gone back and traced it out and it looks like it's this thrift enum decoder in the json module https://github.com/guardian/content-api-models/blob/main/json/src/main/scala/com/gu/contentapi/json/CirceEncoders.scala#L20-L23

PR #241, opened 2024-03-07
Ensure sbt detects version incompatibilities

Divs-B:
This PR https://github.com/guardian/content-api-models/pull/38#discussion_r69474445 set the json's dependency required as "provided" but we are not entirely clearly sure why thta was needed.

ref: https://www.baeldung.com/maven-dependency-scopes#bd-2-provided

It caused problem now because we are not able to see sbt checking on incompatible version, thanks to work done in gha-scala-release where we expect incompatible version fatal errors - https://github.com/guardian/gha-scala-library-release-workflow/blob/add-configuration-guidence/docs/configuration.md#recommended-sbt-plugins

co-authored with @rtyley along with @emdash-ie 

## How to test

Try change version for capi-model-scala-client to older version so that while running test/compile sbt should point out the issues in incomptaible versions.

warnings/errors should be on sbt terminal

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,7 +108,7 @@ lazy val scala = Project(id = "content-api-models-scala", base = file("scala"))
   * JSON parser project
   */
 lazy val json = Project(id = "content-api-models-json", base = file("json"))
-  .dependsOn(scala % "provided")
+  .dependsOn(scala)
   .settings(artifactProductionSettings)
   .settings(
     description := "Json parser for the Guardian's Content API models",


emdash-ie (approved):
I do still wonder whether there’ll be some problem with this that we haven’t anticipated, but hopefully it’ll become obvious quickly if so.

emdash-ie:
(Maybe worth noting for posterity that the specific problem we had was that our dependency tree in `apple-news` contained `content-api-models-scala` and `content-api-models-json` at incompatible versions, so decoding a thrift struct from json caused an error because the apply method didn’t have the number of arguments expected by the decoder – a new field had been added in the newer version used by `content-api-models-scala`. We were confused by this, and hope that removing `"provided"` here will cause an error to be thrown in these situations which makes it clear what’s going wrong.)

Divs-B:
@emdash-ie and I have decided to go ahead to merge this PR, 
excluding "provided" should force sbt to check incompatibly errors in `content-api-models-json` in compile time itself.

PR #244, opened 2024-03-13
Tweak a couple of recipe fields to be arrays of strings instead of strings

frederickobrien:
## What does this change?

Fixing a silly mistake with the thrift. Although it's usually a no-no to change the shape of thrift fields in this case we should be in the clear given it's not being used anywhere yet

## Accessibility

### DIFF
--- a/models/src/main/thrift/content/schema_org.thrift
+++ b/models/src/main/thrift/content/schema_org.thrift
@@ -22,8 +22,8 @@ struct SchemaRecipe {
     14: optional string cookTime
     15: optional string totalTime
     16: optional AuthorInfo author
-    17: optional string suitableForDiet
-    18: optional string cookingMethod
+    17: optional list<string> suitableForDiet
+    18: optional list<string> cookingMethod
 }
 
 struct RecipeStep {


models/src/main/thrift/content/schema_org.thrift
```diff
@@ -22,8 +22,8 @@ struct SchemaRecipe {
     14: optional string cookTime
     15: optional string totalTime
     16: optional AuthorInfo author
-    17: optional string suitableForDiet
-    18: optional string cookingMethod
```
emdash-ie, line 26: These are now backwards-incompatible changes, right? (Assuming you made a release with your previous changes.) Should we deprecate these fields and add new ones, instead?

frederickobrien: Having chatted with @fredex42 (who confirmed with @paulmr) we can get away with it in this case as it's not being used anywhere yet

emdash-ie: I know there shouldn’t be any existing data using these fields with these types, but maybe in the future someone might upgrade to the version we’ve just released that includes the incorrect types?

fredex42: @emdash-ie The thing is, Concierge will never actually generate data with this schema.  So, if somebody does accidentally update to this version then they will get a 500 error if there is any (list) data in these fields.  

That won't be an issue unless they have explicitly requested schema.org content with the `schemaOrg=true` parameter.

The difference is that Concierge will _never_ have generated data in this form, so I think it's right to update the library asap before the incorrect version has the chance to spread.

PR #245, opened 2024-03-20
Timeline model

davidfurey:
## Accessibility

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -903,18 +903,20 @@ struct ListElementFields {
     2: optional ListType type;
 }
 
-struct TimelineItem {
-    1: required list<BlockElement> elements = [];
+struct TimelineEvent {
+    1: required list<BlockElement> body = [];
 
-    2: optional string title;
+    2: optional BlockElement main;
 
-    3: optional string date;
+    3: optional string title;
 
-    4: optional string label; // should this be a list?
+    4: optional string date;
+
+    5: optional string label;
 }
 
 struct TimelineSection {
-    1: required list<TimelineItem> items;
+    1: required list<TimelineEvent> events;
 
     2: optional string title;
 }
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -148,6 +148,8 @@ enum ElementType {
     RECIPE = 22
 
     LIST = 23
+
+    TIMELINE = 24
 }
 
 enum TagType {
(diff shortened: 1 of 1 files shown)


rhystmills (approved):
Looks consistent with our requirements 👍

models/src/main/thrift/content/v1.thrift
```diff
@@ -901,6 +903,28 @@ struct ListElementFields {
     2: optional ListType type;
 }
 
+struct TimelineEvent {
+    1: required list<BlockElement> body = [];
+
+    2: optional BlockElement main;
```
jonathonherbert, line 909: What's the difference between a main element and the body elements?

models/src/main/thrift/content/v1.thrift
```diff
@@ -901,6 +903,28 @@ struct ListElementFields {
...
+
+    3: optional string title;
+
+    4: optional string date;
+
+    5: optional string label;
```
jonathonherbert, line 915: How is this label used? We already have a title, but it looks like there's additional description required, I wonder whether we can document why here

models/src/main/thrift/content/v1.thrift
```diff
@@ -901,6 +903,28 @@ struct ListElementFields {
...
 }
 
+struct TimelineEvent {
+    1: required list<BlockElement> body = [];
+
+    2: optional BlockElement main;
+
+    3: optional string title;
+
+    4: optional string date;
```
jonathonherbert, line 913: Presume the reason we are not using `CapiDateTime` here is because this is audience facing and may be formatted in many different ways?

jonathonherbert (approved):
These changes all look reasonable. Thrift being the authority for so much of our modelling, I do feel like a lot of what's here would benefit from more description, especially when so many modelling decisions we make are quite tactical.

I've added some comments asking questions about what are likely perfectly reasonable modelling decisions, and I wonder if in answering them you could add those comments to the model for future reference.

models/src/main/thrift/content/v1.thrift
davidfurey: Similar to the `main` block (used for main media) and `body` blocks (used for the article), we will allow a single main element and then have the rest of the content in body. This will make it easier for the platforms to render the first image differently.

models/src/main/thrift/content/v1.thrift
davidfurey: In this example "Primary" is a label. 
Screenshot 2024-03-28 at 15 54 21

models/src/main/thrift/content/v1.thrift
davidfurey: We need to allow timelines to include events where the date is not exactly known, where it could be a range, etc. At some point in the future I'd like to explore making this machine readable (perhaps using https://www.loc.gov/standards/datetime/). But for version 1, a string seems reasonable.

PR #247, opened 2024-06-13
Expand list model to include fields required by Mini profiles and Multi-byline formats

simonbyford:
## What does this change?

Expands the list model to include fields required by Mini profiles and Multi-byline formats. This brings the model into line with flexible-model: https://github.com/guardian/flexible-model/pull/69

## Accessibility

### DIFF
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -887,6 +887,7 @@ enum ListType {
     KEY_TAKEAWAYS = 1,
     Q_AND_A_EXPLAINER = 2,
     MINI_PROFILES = 3,
+    MULTI_BYLINE = 4,
 }
 
 struct ListItem {
--- a/models/src/main/thrift/content/v1.thrift
+++ b/models/src/main/thrift/content/v1.thrift
@@ -886,6 +886,7 @@ struct RecipeElementFields {
 enum ListType {
     KEY_TAKEAWAYS = 1,
     Q_AND_A_EXPLAINER = 2,
+    MINI_PROFILES = 3,
 }
 
 struct ListItem {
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Looks good to me!

emdash-ie (approved):
This looks fine to me :)

emdash-ie:
For the record/history, this PR caused us a small bit of trouble! Because https://github.com/guardian/content-api/pull/2879 updated porter [to write the contributors as a list of IDs](https://github.com/guardian/content-api/pull/2879/files#diff-e27431c0338e74556d24ce76d8bdbd64854092cc2c0eb716c0bb5da034033ed7R511), defining the type of those contributors to be a list of tag objects here is effectively imposing a requirement on concierge to transform those ids into tag objects, or drop them. (This is because this model defines the structure of the data that concierge must output based on what it gets from elasticsearch.)

Since we never updated concierge to do this, once we got one of these fields in CODE (while testing https://github.com/guardian/content-api/pull/2925), concierge was unable to handle the field and broke whenever it tried to output the content containing the field. (This also broke crier, since it calls concierge.)

I’ll keep this in mind in future when reviewing PRs like this, and possibly we should require tests of some sort on this repository that would catch this – I’m not sure what those would be though.

It’s also worth noting that since the flexible-model defines the structure of the data accepted by porter, which it then transforms before writing it into elasticsearch, we don’t necessarily want the two models to be aligned: only where porter makes no changes before writing to elasticsearch.

We should also consider updating concierge (and crier?) to continue operating well when they encounter an error like this.

(Something else I’d like to look into is having another model that’s shared between porter and concierge and represents the data written to elasticsearch. To preserve the ability to make changes, we’d need to have tests that new versions of the library can’t write data that breaks old versions of the library, but that seems like a plausible feature. (For example, ophan’s avro-logger has [similar tests](https://github.com/guardian/ophan/blob/5a7475122afecefb9b08a038976e7afb0793b6f1/avro-logger/src/test/scala/lib/AvroSchemaCompatibilityTest.scala) which make sure new schema versions can always read data written by old versions.))

PR #249, opened 2024-07-03
Update thrift library

fredex42:
Updates the thrift library to the most recent (Apr 2024)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,12 +4,12 @@ import sbtrelease.ReleaseStateTransformations.*
 import sbtversionpolicy.withsbtrelease.ReleaseVersion
 
 // dependency versions
-val contentEntityVersion = "3.0.3"
-val contentAtomVersion = "4.0.4"
+val contentEntityVersion = "4.0.0"
+val contentAtomVersion = "5.0.0"
 val storyPackageVersion = "2.2.0"
 val thriftVersion = "0.20.0"
 val scroogeVersion = "22.1.0" // update plugins too if this version changes
-val circeVersion = "0.14.1"
+val circeVersion = "0.14.9"
 val fezziwigVersion = "2.0.0"
 
 // dependency versions (for tests only)
@@ -117,7 +117,7 @@ lazy val json = Project(id = "content-api-models-json", base = file("json"))
       "io.circe" %% "circe-core" % circeVersion,
       "io.circe" %% "circe-generic" % circeVersion,
       "io.circe" %% "circe-parser" % circeVersion,
-      "io.circe" %% "circe-optics" % circeVersion,
+      "io.circe" %% "circe-optics" % "0.14.1",  //circe-optics is now released at a difference cadence to circe main project - https://github.com/circe/circe-optics
       "org.scalatest" %% "scalatest" % scalaTestVersion % Test,
       "com.google.guava" % "guava" % guavaVersion % Test,
       "org.gnieh" %% "diffson-circe" % diffsonVersion % Test
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ import sbtversionpolicy.withsbtrelease.ReleaseVersion
 val contentEntityVersion = "3.0.3"
 val contentAtomVersion = "4.0.4"
 val storyPackageVersion = "2.2.0"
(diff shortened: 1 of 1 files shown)


marjisound:
Hi @fredex42  is there a plan to release this PR any time soon? I'm mostly interested in the libthrift version update

fredex42:
Hi @marjisound thanks for the bump. I don't see why not though it was a while since I did this! Seems to be missing a review so maybe @JustinPinner @jonathonherbert amd/or @rtyley could give it the once-over? Am on Feast now so much further away from this stuff.

JustinPinner:
IIRC we had some issues with reserved keywords in later thrift versions, but if this version has been integrated and tested successfully then we should be good to go. Of course I could be crossing this problem with something else. Will have to look back at where and when we had issues.

marjisound:
> IIRC we had some issues with reserved keywords in later thrift versions, but if this version has been integrated and tested successfully then we should be good to go. Of course I could be crossing this problem with something else. Will have to look back at where and when we had issues.

Thanks @JustinPinner I'm not sure about thrift but I also noticed that the scrooge is banning reserved keywords which caused issue in this #231 @rtyley 's comment: https://github.com/guardian/content-api-models/pull/231#issuecomment-1900668948

JustinPinner:
My apologies @marjisound, that's probably what I was (mis)remembering. In which case we should be ok with this version. Will approve it now.

JustinPinner (approved):
I think this should be safe to go ahead 👍

rtyley (commented):
> IIRC we had some issues with reserved keywords in later thrift versions, 

I think we're talking about `content-api-models` Thrift definitions using the field name `end`, which was [reserved](https://github.com/twitter/scrooge/issues/259) in later versions of Thrift, right?

https://github.com/guardian/content-api-models/blob/86752f2d1af7607d0b662939ddbec58aa94eba9f/models/src/main/thrift/content/v1.thrift#L393-L394

Mitigations we've used against this issue have varied over time:

- using `scroogeDisableStrict`, which eventually [stopped working](https://github.com/twitter/scrooge/issues/259#issuecomment-1900743695)
-  downgrading to old versions of the Thrift library or Scrooge library

I now have the shadowy recollection that the _newest_ versions of Thrift have **reversed their decision** to make so many words 'reserved' - they decided that reserving _all_ the reserved words from 20 or so different supported programming languages was too much. So now, so long as the language you're using doesn't reserve that word, and it's not on a shorter list of universally reserved words, it should be ok. But I think I only _heard_ that, and I don't know what version of Thrift that came in with.

> but if this version has been integrated and tested successfully then we should be good to go.

Just checking - is that the case? @fredex42 made a pre-release artifact for this PR in July 2024 as `26.0.0-PREVIEW.agupdate-thrift.2024-07-04T1251.0144ace4`. Note that a different set of changes has already been released as [`v26.0.0`](https://github.com/guardian/content-api-models/releases/tag/v26.0.0). Was that pre-release artifact tried out?

It might be worth rebasing this PR on latest main, doing a new [pre-release ](https://github.com/guardian/gha-scala-library-release-workflow/blob/main/docs/making-a-release.md#choose-release-type) and seeing if that compiles/passes tests in a couple of test projects like `content-api` or whatever.

I could be being over-cautious here, if everyone else is ok with what we've got it may well be fine.

emdash-ie:
> I could be being over-cautious here, if everyone else is ok with what we've got it may well be fine.

I also feel the need for this caution

rtyley:
> Interestingly, I believe the content-api-models-json subproject is already using v0.19.0 because the 0.15.0 dependency is evicted by [fezziwig's 0.19.0 dependency](https://github.com/guardian/fezziwig/blob/3b2d3340a7864ad17a157f750014cde3df4a1aaa/build.sbt#L32).

This is a really good point - after checking Ophan's code, it appears that it is already (unintentionally, maybe?!) [using libthrift 0.20.0](https://github.com/guardian/ophan/issues/5924#issuecomment-2590491317) so we can disregard the Ophan concern for evaluating this PR ✨

marjisound:
Hi all, just checking in to see if there are any updates on testing this change and whether it’s likely to be merged soon

PR #250, opened 2024-07-24
Add special crossword type

andrew-nowak:
## What does this change?

Adds the upcoming Special crossword type to the CAPI models. Required to have the crossword type correctly parsed from ES and returned in the Concierge response.


JustinPinner (approved):
Looks entirely reasonable to me 👍

PR #269, opened 2025-10-02
Rename `end` field that uses reserved Thrift word

rtyley:


emdash-ie:
> We should check if this field name is also used in thrift in flexible-model, and whether we need to make a similar change there.

It looks like [the same field name exists in flexible-model](https://github.com/guardian/flexible-model/blob/6969333d0bf883eaa03fcf4ae61e3769724954c8/src/main/thrift/content.thrift#L421) in ElementFields and MembershipElementFields. What I’m not sure of is how exactly these thrift definitions are used: are they used for any serialisation/deserialisation that uses the field names?

PR #307, opened 2026-06-04
Add some thrift round-trip tests

emdash-ie:
## What does this change?

The tests on this repository (in the json project) test the derived circe decoders thoroughly, but there’s not much testing of the thrift binary protocol output. Having these tests is a starting point that should give us more confidence in PRs on this repository.


scala/src/test/scala/com.gu.contentapi.scala/ThriftRoundTripSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.scalatest.Assertion
+import com.twitter.scrooge.ThriftStruct
+
+class ThriftRoundTripSpec extends FlatSpec with Matchers {
+  it should "round-trip an ItemResponse" in {
+    checkRoundTrip(
+      Path.of("hot-divorcee-summer.binary.thrift"),
+      ItemResponse,
+      (item: ItemResponse) => {
+        item.content.get.fields.get.headline shouldBe Some("Hot divorcee summer: get ready for big hats, hot sex and don’t-care energy")
```
jonathonherbert, line 24: this is an incredible choice for a fixture and I absolutely approve

jonathonherbert (approved):
> At the moment, if there’s a PR open on this repository I don’t feel comfortable merging it based on the CI because I’m not confident that the CI will catch any breakages. Instead I feel like I should make a preview release from the PR branch, depend on that in porter, and manually test porter (or rely on porter’s CI, possibly) – this is a lot of effort. Adding these tests to CI is my attempt to automate that manual testing, by copying what porter does.
> 
> Aside from the specific perspective of automating that testing, I think adding tests that use the code provided by this library in ways we expect consumers of the library to use it seems like a reasonable way to make sure changes to the library aren’t breaking things for consumers without us noticing?

Yes, I hadn't thought about this in the context of dependency upgrades, etc. — and that we can e.g. merge dependabot PRs more confidently with this test. 👍  Thanks for explaining!

PR #309, opened 2026-06-17
Add fallbackContent to InteractiveElementFields and add ResponsiveImageElement types

emma-imber:
## What does this change?
Adds a `fallbackContent` field to interactive elements, and a new `ResponsiveImage` element. This will enable us to add images as fallback content for interactives that can't be displayed in Apple News. See the RFC here: https://github.com/guardian/flexible-model/discussions/98

## How has this change been tested?
We've tested this on CODE-AARDVARK and verified that a responsive image used as fallback content is correctly returned by CAPI.


jonathonherbert (approved):
Tested end-to-end with flexible-content and Concierge, can see valid output for interactive elements 👍 

Image

PR #310, opened 2026-06-19
Bump to content-atom v12.0.1

davidfurey:
This adds a new optional parameter codecs for media atoms. See https://github.com/guardian/content-atom/pull/216


waisingyiu (approved):
👍

PR #314, opened 2026-07-09
Add thriftToFile helper for creating tests

emdash-ie:
## What does this change?

This PR adds a helper function that makes it easier to add new tests to the ThriftRoundTripSpec, with some helpful info in scaladoc. It was written as part of adding a test to https://github.com/guardian/content-api-models/pull/311.

## How has this change been tested?

We used it as described for writing the test in https://github.com/guardian/content-api-models/pull/311.


andrewHEguardian (approved):
Fantastic! I think I would have been able to add the test myself if we'd had this helper in place

PR #315, opened 2026-07-09
Add tests of enum serialisation

emdash-ie:
## What does this change?

We kind of test this already with the ItemResponse test, but it’s worth adding extra ones since we’re specifically worried. I generated the binary values with the thriftToFile helper on thrift v0.15.0 (by downgrading and depending on a local version of fezziwig which was also downgraded), so that we can be confident updating to latest thrift hasn’t changed anything.

This should give us extra confidence for https://github.com/guardian/articles-and-publishing/issues/517


groakland (commented):
The expected TagType values are currently discarded (_tagType), so this test only verifies that each fixture can be decoded and re-encoded to the same bytes. It does not confirm that the fixture decodes to the intended enum value.

We could pass an assertion to checkRoundTrip for each entry, e.g. asserting that the decoded Tag.type equals the expected TagType? That would ensure the test catches an enum mapping or wire-encoding issue rather than only validating a byte-preserving round trip.

Why this matters:
A round-trip can pass even if decoding maps an enum to the wrong value, because the same incorrect value may be serialized back into the original bytes. Asserting the decoded TagType makes the test validate the actual semantic mapping that this PR is intended to protect.

groakland (approved):
LGTM with the latest changes. Approved

PR #319, opened 2026-07-16
Explain tests and their guarantees in readme

emdash-ie:
## What does this change?

For people who are new to the project, it’s good to know what tests there are so they know how much confidence to take from passing CI on their PRs. This PR adds explanation in the readme of what tests there are and what they guarantee (and what their limitations are).


jonathonherbert (approved):
Useful docs, thank you!

